PR #10, opened 2019-12-02
python code to generate reports and graphs given ADM mart data

shishehchi:
two python classes. One class reports on models including historical performance. The other class reports on the latest model and their predictors


operdeck:
We should include this in the CI process - we try do so for all Pega Github projects. For the R package I'm doing that already, see the Travis configuration. Preferably that should also include some code coverage metric.

PR #11, opened 2020-01-15
extra methods added to explore model and predictor data

shishehchi:
added a few more methods to the already existing classes to generate more model and predictor report graphs


shishehchi:
I will cancel this pull request due to conflicts. Something was set up wrong. Will merge again

PR #14, opened 2020-10-23
IH analysis

shishehchi:
added a py file that contains a number of functions to explore IH data. Also added a notebook file that uses the IH analysis functions and runs it on an example IH dataset


operdeck (approved):

operdeck:
Consider adding some doc either as readme or as github pages

PR #25, opened 2021-10-12
Plot Control and Test groups CTR for the action/treatment combination

grishasen:
# Calculate z-score and visualize significance


Grinders:
Will create another pull request

Issue #30, opened 2021-11-12
Align python & R more


operdeck:
For easy of use, we should align both a little more.

- Datamart read follows a different style. The R version assumes the standard datamart names, then chops off the py/px prefixes and does some capitalization, so we get fields like ResponseCount not pyresponsecount. The Python version requires a mapping to the database field names. 
- Basic plotting functions on datamart data should be aligned more between the two. Both have some standard plotting functions but naming and functionality is different. Functionality doesn't need to be 100% the same, we're not building a production toolkit but somewhat more aligned would be useful.
- For R we build a package, and the process includes unit tests, coverage, documentation and help and continuous integration (via Github Actions). For Python there is no such thing and we require a clone of the source files.
- Utilities like PMML conversion etc can continue live in just R - they're not used much anyway

StijnKas:
Agreed, I've already started work on the first point. Current idea is to set a default mapping for the common variables, which you can overwrite with a custom mapping (and any additional variables you supply will then also be included). 

For the second point, it's probably good to have a general 'ADM data' class which takes care of the data imports, preprocessing and combination of various files and then have a separate visualisation class which takes in that standardised data to make the various visualisations (that way we can also easily substitute different plotting libraries).

operdeck:
Hey, sounds good. From experience I'm just a little wary of abstraction
layers. Settling to just plotly (for Python) seems good enough to me.

Op vr 12 nov. 2021 11:52 schreef StijnKas ***@***.***>:

StijnKas:
Implemented in Python in PR #32

StijnKas:
With #32 we can probably close this issue. It's not fully aligned, but the function names are now unified and the features are renamed the same way. Packaging, unit tests and integration testing are currently not implemented, but that's not so much aligning as implementing those as features anyways. Introduced #36 and #37 so it doesn't fall off the radar.

Issue #29, opened 2021-11-12
Use arrow also for internal reading in readDSExport R function


operdeck:
Consider arrow package for fast reading (& writing perhaps) of LDJSON files and even ordinary csv.

StijnKas:
Implemented in Python in PR #32

operdeck:
Using it also in ADMDatamart function in R. TODO use it also when reading the data.json from the export zips.

operdeck:
Arrow already used by R's ADMDatamart but to be used internally also when reading the data.json file from a DS export

PR #32, opened 2021-12-15
Python refactoring

StijnKas:


StijnKas:
Implements #30 and #29

PR #59, opened 2022-11-30
Caching

StijnKas:
This PR adds caching of tables, currently implemented in the ADMDatamart class. Simply call `save_data(path)` which will save the model data and predictor data as .feather files to the path supplied in the function. The default scanning table names in ADMDatamart now also recognise these cached files.

Also, I changed the default importing behaviour by using Polars rather than PyArrow, and added logging during that importing function in `cdh_utils`.


StijnKas:
Running into some Polars issues, namely the fact that it can't handle Pega's OOTB scientific notation format when reading from json. I've filed an issue there - as reading with polars is quite a bit faster than the previous pandas-based method. So I think i'll wait for that to be resolved before continuing with this PR, especially since these are not essential changes.

StijnKas:
Underlying issue seems to be fixed in the backend json parser lib, so once that's propagated to Polars I can continue this work. Will probably be a few days still.

StijnKas:
Decided to split this up into two separate PRs: https://github.com/pegasystems/pega-datascientist-tools/pull/63 & https://github.com/pegasystems/pega-datascientist-tools/pull/61, since https://github.com/pegasystems/pega-datascientist-tools/pull/60 was causing too many merge conflicts. All items here have been included in those.

PR #64, opened 2023-01-11
Update cdh_utils.py

yusufuyanik1:
Testing


python/pdstools/utils/cdh_utils.py
```diff
@@ -723,3 +723,31 @@ def getToken(credentialFile, verify=True, **kwargs):
         auth=(creds["Client ID"], creds["Client Secret"]),
         verify=verify,
     ).json()["access_token"]
+
+
+def calc_reach(num_positives):
```
StijnKas, line 728: Add docstrings

PR #65, opened 2023-01-16
Interaction history based model performance

Grinders:
The goal of this notebook is to provide a high-level review of the model performance  and metrics (Accuracy, F1, ROC AUC etc.) based on interaction history data.

### DIFF
--- a/examples/ih/ih_model_review_sample.Rmd
+++ b/examples/ih/ih_model_review_sample.Rmd
@@ -1,0 +1,1179 @@
+---
+title: "`r params$title`"
+subtitle: "`r params$subtitle`"
+author: "Pega Data Scientist's Tools"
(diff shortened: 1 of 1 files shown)


Grinders:
Proposing a notebook based on some reports used in customer environment.

StijnKas:
Hey @Grinders, thanks for the notebook, looks like an interesting analysis. Just out of interest - why would you redefine some of the functions at the top (fromPRPCDateTime, safe_range_auc, pr_auc_from_probs)? They should already be in pdstools natively, so for keeping things up to date I'd prefer to use those directly. Did you need to adapt them to fit a different need somehow?

operdeck:
To be reconsidered later along with other IH style analyses

PR #73, opened 2023-03-14
pdstools V3: Polars backend, Python healthcheck

StijnKas:
This PR contains a lot of changes, many of them breaking, and many of them introducing new features. The full changes will be summarized in the release notes of the new tag, but as a summary:

- All files now use Polars as the back-end, replacing Pandas nearly completely
- Matplotlib support is dropped, Plotly is now the only plotting backend
- The healthcheck from R is now ported over fully to Python, thanks to @yusufuyanik1


sushantkhare:
Hi team, not sure if this is the right place for this comment but I am facing issues when I am trying to use the ADMDatamart function for the exports from my local machine. I am running out of memory and getting this error: **MemoryError: Unable to allocate output buffer.**

For reference I am just trying to get a datamart object as I need to look at some quick metrics. I have been able to export json file using readDSExport but as you know it provides raw data so it will take me quite some time to actually reach the analysis I am trying to do. 

Hence, I tried using **_dm = ADMDatamart('my-path', , import_strategy='lazy')_** where my-path has string value of the directory where I have the zipped exports. 

I was hoping to use dm t explore the pre-made functions of pdstools. Maybe you have encountered this issue before? Are there are any pointers? Similar memory issue while trying to read the json file got resolved as readDSExport uses polars. Not sure what I could do here.

Please let me know if you would need more information from me, thanks!

PR #77, opened 2023-03-22
Streamlit

yusufuyanik1:
- Removed pdf and word export options from streamlit app. I believe they are not useful (you can't see the plots appropriately) and if we want to keep them, we have to make sure they work properly (Ex. pdf export doesn't have the _download file_ option and it generates the pds file it in wherever you run the app.)

- If the user feeds a directory path with non-standard file names to ADMDatamart, it would return a NONE object and result in an error like (Nonetype object has no modeldata ...). But I think it should raise a more verbose error. To achieve that:
added _is_url function to test if the given path is a URL. To make sure before raising errors such as: Not a directory or this directory doesn't have files with standard standard file names(see admdatamart changes)

- Added HealthCheckModel.qmd to create a healthcheck in case only the modeldata is uploaded (currently if filepath option is used, it wouldnt work), only works with upload file.

- updated the extract_keys argument, it was depriciated in the app.

- --server.maxUploadSize default is set to 2GB

Some controversial changes with removing pds, word options and I am not sure if we should change the error messages of ADM. Maybe sticking to st.warning() messages in streamlit is a better idea, waiting for your comments

### DIFF
--- a/python/pdstools/adm/ADMDatamart.py
+++ b/python/pdstools/adm/ADMDatamart.py
@@ -286,10 +286,20 @@ class ADMDatamart(Plots):
                 df2 = df2.join(df1.select(pl.col("ModelID").unique()), on="ModelID")
 
         if self.import_strategy == "eager":
-            if df1 is not None:
-                df1 = df1.collect().lazy()
-            if df2 is not None:
-                df2 = df2.collect().lazy()
+            try:
+                if df1 is not None:
+                    df1 = df1.collect().lazy()
+                if df2 is not None:
+                    df2 = df2.collect().lazy()
+            except pl.ComputeError as e:
+                raise Exception(
+                    """Unexpected Time format! 
+                        Please refer to :meth:`._set_types` to change datetime format \n
+                        If you are using HealthCheck app, use `Edit Parameters` button above
+                        and enter the date format in your files. \n
+                        You can check https://strftime.org/ for formatting options
+                        """
+                ) from e
         return df1, df2
 
     def _import_utils(
--- a/python/pdstools/app/datamart_utils.py
+++ b/python/pdstools/app/datamart_utils.py
@@ -29,12 +29,7 @@ def remove_duplicate_expressions(exprs):
 def import_data(params, default=0, **kwargs):
     config_type = st.selectbox(
         "Data import type",
-        options=[
-            "Sample dataset",
-            "Filepath",
-            "File upload",
-            "S3 Bucket",
-        ],
+        options=["Sample dataset", "Filepath", "File upload", "S3 Bucket"],
         index=default,
     )
     if config_type == "File upload":
@@ -48,7 +43,7 @@ def import_data(params, default=0, **kwargs):
             )
             params["kwargs"] = ADMDatamart_options()
             if st.checkbox("Use this data source"):
-                if model_file is not None:
+                if model_file and predictor_file is not None:
                     data = import_datamart(
                         ADMDatamart,
                         model_df=model_file,
@@ -56,7 +51,10 @@ def import_data(params, default=0, **kwargs):
                         **params["kwargs"],
                     )
                 else:
-                    "Please upload your datamart files."
+                    st.warning(
+                        """Please upload both Model Snapshot and 
+                               Predictor Binning Snapshot."""
+                    )
     elif config_type == "Sample dataset":
         data = import_datamart(datasets.CDHSample)
     elif config_type == "Filepath":
@@ -84,8 +82,6 @@ def import_data(params, default=0, **kwargs):
                     "pyName",
                     "pySnapshotTime",
                 ]
-                import sys
-
                 try:
                     from boto3 import client
                 except ModuleNotFoundError:
@@ -123,6 +119,15 @@ def import_data(params, default=0, **kwargs):
         return params, data
     except UnboundLocalError:
         return params, None
+    except pl.ComputeError:
+        st.write(
+            """Unexpected Time format! 
+                 Please use Edit Parameters button above and enter the date format in 
+                 your files. You can check https://strftime.org/ for formatting
+                 options
+                 """
+        )
+        return None, None
 
 
 @st.cache_resource(show_spinner=True)
@@ -138,7 +143,7 @@ def ADMDatamart_options():
     if st.checkbox("Edit Parameters"):
         extract_treatment = st.checkbox("Extract treatments", False)
         if extract_treatment:
-            params["extract_treatment"] = "pyName"
+            params["extract_keys"] = "pyName"
         context_keys = ["Issue", "Group", "Channel", "Direction"]
         if extract_treatment:
             context_keys.append("Treatment")
--- a/python/pdstools/app/datamart_utils.py
+++ b/python/pdstools/app/datamart_utils.py
@@ -43,17 +43,20 @@ def import_data(params, default=0, **kwargs):
             )
             params["kwargs"] = ADMDatamart_options()
             if st.checkbox("Use this data source"):
-                if model_file and predictor_file is not None:
+                if predictor_file is None and model_file is not None:
+                    data = import_datamart(
+                        ADMDatamart,
+                        model_df=model_file,
+                        predictor_df=predictor_file,
+                        predictor_filename=None,
+                        **params["kwargs"],
+                    )
+                elif model_file and predictor_file is not None:
                     data = import_datamart(
                         ADMDatamart,
                         model_df=model_file,
                         predictor_df=predictor_file,
                         **params["kwargs"],
-                    )
-                else:
-                    st.warning(
-                        """Please upload both Model Snapshot and 
-                               Predictor Binning Snapshot."""
                     )
     elif config_type == "Sample dataset":
         data = import_datamart(datasets.CDHSample)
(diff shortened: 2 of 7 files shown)


python/pdstools/adm/ADMDatamart.py
```diff
@@ -286,10 +286,20 @@ def import_data(
...
-            if df1 is not None:
-                df1 = df1.collect().lazy()
-            if df2 is not None:
-                df2 = df2.collect().lazy()
+            try:
+                if df1 is not None:
+                    df1 = df1.collect().lazy()
+                if df2 is not None:
+                    df2 = df2.collect().lazy()
+            except pl.ComputeError as e:
```
StijnKas, line 294: I don't think this'd be the right way of doing it. This is just a catch-all exception catcher (ComputeError is very generic) and it doesn't actually _have_ to mean it's something to do with the timestamps. How to catch that though, I'm not sure.

Also, I wouldn't raise a general Exception but I'd try to raise a more specific error. Datetime itself raises a ValueError, so that's already better.

python/pdstools/app/datamart_utils.py
```diff
@@ -48,15 +43,21 @@ def import_data(params, default=0, **kwargs):
...
-                if model_file is not None:
+                if predictor_file is None and model_file is not None:
+                    data = import_datamart(
+                        ADMDatamart,
+                        model_df=model_file,
+                        predictor_df=predictor_file,
+                        predictor_filename=None,
+                        **params["kwargs"],
+                    )
+                elif model_file and predictor_file is not None:
```
StijnKas, line 54: This looks like a weird expression to me - are you trying to check if 'model_file' is True?

python/pdstools/app/datamart_utils.py
```diff
@@ -48,15 +43,21 @@ def import_data(params, default=0, **kwargs):
             )
             params["kwargs"] = ADMDatamart_options()
             if st.checkbox("Use this data source"):
-                if model_file is not None:
+                if predictor_file is None and model_file is not None:
+                    data = import_datamart(
+                        ADMDatamart,
+                        model_df=model_file,
+                        predictor_df=predictor_file,
```
StijnKas, line 50: If you give `predictor_filename = None`, then please don't pass a `predictor_df`. filenames take priority. So, if you want to _not_ pass a predictordata at all, just pass `predictor_filename = None` and no `predictor_df`.

python/pdstools/app/datamart_utils.py
```diff
@@ -129,7 +128,8 @@ def get_from_s3(bucket, key, sql_query):
 def import_datamart(_fun, *args, **kwargs):
     dm = _fun(*args, **kwargs)
     dm.modelData = dm.modelData.collect()
-    dm.predictorData = dm.predictorData.collect()
+    if kwargs.get("predictor_df") is not None:
```
StijnKas, line 131: Do we need to do this? I'd rather not

python/pdstools/app/datamart_utils.py
```diff
@@ -138,7 +138,7 @@ def ADMDatamart_options():
     if st.checkbox("Edit Parameters"):
         extract_treatment = st.checkbox("Extract treatments", False)
         if extract_treatment:
-            params["extract_treatment"] = "pyName"
+            params["extract_keys"] = "pyName"
```
StijnKas, line 141: Extract_keys is now a boolean, not a string value

python/pdstools/utils/cdh_utils.py
```diff
@@ -137,6 +134,30 @@ def readDSExport(
     return import_file(file, extension, **reading_opts)
 
 
+def _is_url(path: str, filename: str):
```
StijnKas, line 137: Please check for response.status_code == 200. That makes sure it's a _valid and working_ URL

python/pdstools/utils/cdh_utils.py
```diff
@@ -266,20 +280,30 @@ def get_latest_file(path: str, target: str, verbose: bool = False) -> str:
...
 
     supported = [".json", ".csv", ".zip", ".parquet", ".feather", ".ipc", ".arrow"]
+    try:
+        files_dir = [
+            f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))
+        ]
+        files_dir = [
+            f for f in files_dir if os.path.splitext(f)[-1].lower() in supported
+        ]
+    except FileNotFoundError as e:
```
StijnKas, line 290: Is this right? FileNotFound doesn't sound to me like it's a directory not found?

python/pdstools/utils/cdh_utils.py
```diff
@@ -325,6 +349,14 @@ def getMatches(files_dir, target):
         match = [file for name in names if re.findall(name.casefold(), file.casefold())]
         if len(match) > 0:
             matches.append(match[0])
+    if len(matches) == 0:
+        raise FileNotFoundError(
+            """Couldn't find a file with default ADM table names. Please try feeding model and predictor files seperately. \n
+        
+        Or check out getMatches function at https://github.com/pegasystems/pega-datascientist-tools/blob/master/python/pdstools/utils/cdh_utils.py
```
StijnKas, line 356: I'd rather not point to source code - would rather include those default names in the function docstring so they can be included in the docs

python/pdstools/utils/cdh_utils.py
```diff
@@ -325,6 +349,14 @@ def getMatches(files_dir, target):
         match = [file for name in names if re.findall(name.casefold(), file.casefold())]
         if len(match) > 0:
             matches.append(match[0])
+    if len(matches) == 0:
+        raise FileNotFoundError(
+            """Couldn't find a file with default ADM table names. Please try feeding model and predictor files seperately. \n
```
StijnKas, line 354: I'd just fstring in the given file path here, just so it's explicit _where_ we couldn't find the files.

python/pdstools/utils/cdh_utils.py
```diff
@@ -648,14 +680,7 @@ def _capitalize(fields: list) -> list:
...
 def _polarsCapitalize(df: pl.LazyFrame):
-    return df.rename(
-        dict(
-            zip(
-                df.columns,
-                _capitalize(df.columns),
-            )
-        )
-    )
+    return df.rename(dict(zip(df.columns, _capitalize(df.columns))))
```
StijnKas, line 683: I think you have a slightly different linter than I do :) I don't mind it pulling this back to one line, but I do mind it pingponging back and forth between every single PR. Let's align on that soon

python/tests/test_cdh_utils.py
```diff
@@ -58,7 +58,7 @@ def test_find_default_predictors():
 
 
 def test_file_not_found():
-    with pytest.raises(FileNotFoundError):
+    with pytest.raises(Exception):
```
StijnKas, line 61: Again, I don't like general Exceptions. Let's try to specify those. FileNotFoundError is much nicer, imo.

python/tests/test_cdh_utils.py
```diff
@@ -196,11 +190,12 @@ def test_cdh_sample_autodiscovered_locally():
 
 
 def test_file_not_found_in_good_dir():
-    assert cdh_utils.readDSExport(path="data", filename="models") == None
+    with pytest.raises(ValueError):
```
StijnKas, line 193: I explicitly didn't want this to raise an error, by the way. The desired behaviour was for this _not_ to bug, but instead to return None. That way you don't have to catch that in the ADMDatamart class - and it's pretty obvious that if readDSExport didn't return anything, it didn't find the file

python/tests/test_cdh_utils.py
```diff
@@ -196,11 +190,12 @@ def test_cdh_sample_autodiscovered_locally():
...
+        cdh_utils.readDSExport(path="data", filename="models")
 
 
 def test_file_not_found_in_bad_dir():
-    with pytest.raises(FileNotFoundError):
+    with pytest.raises(Exception):
```
StijnKas, line 198: See above

StijnKas (changes requested):

StijnKas:
So, your top two points I didn't really address in the review:
- Can we try to at least make pdf work? I know word probably won't be a very popular export option, but if we just set plotly to a static plot type (can even use the parameters section for that) it'd be really nice if we could at least support it. 
- Not sure to be honest. I agree it would be nice if we could give more verbose pointers, but I wanted the OOTB functionality to simply return None from the datamart. Perhaps this could be a keyword argument or something. Or perhaps we could rearrange things and check for getMatches in the ADMDatamart class ahead of time?

python/pdstools/app/datamart_utils.py
StijnKas: I mean I'd rather not collect predictordata at all :D

python/pdstools/utils/cdh_utils.py
StijnKas: Right. So what I mean is you raise a FileNotFoundError, but in the text you mention that the directory does not exist. That's typically not very verbose - and if I got a FileNotFoundError and the text mentioned that the directory does not exist - I wouldn't know what to believe

python/pdstools/utils/cdh_utils.py
yusufuyanik1: I wrote a docstring and added the default names to getMatches function now. But in the app you won't be able to see the docstring. So it would be nice if we could point the user to somewhere.

python/pdstools/app/datamart_utils.py
yusufuyanik1: `@st.cache_resource(show_spinner=True)` caching was giving an error for LazyFrame but you already created a custom caching function for lazyframes I believe. Do we have it already in the repo?

StijnKas: @st.cache(
    hash_funcs={pl.LazyFrame: lambda x: hash(x.explain(optimized=False))},
)

PR #78, opened 2023-03-23
ADMExplained Article

yusufuyanik1:


python/pdstools/adm/ADMDatamart.py
```diff
@@ -343,7 +343,7 @@ def _import_utils(
             df = cdh_utils.readDSExport(
                 filename=name, path=path, verbose=self.verbose, **reading_opts
             )
-        elif isinstance(name, pl.DataFrame) or isinstance(name, pl.LazyFrame):
+        if isinstance(name, pl.DataFrame) or isinstance(name, pl.LazyFrame):
```
StijnKas, line 346: Oops. Accidentally pushed this along with it. I think this is desired anyway, but wasn't my intention.

PR #83, opened 2023-03-30
Model-based Health Check

yusufuyanik1:
- Added HealthCheckModel.qmd along with changes to related files to integrate model-based Health Check run
- Increases HealthCheck app file upload size to 2GB


python/pdstools/adm/ADMDatamart.py
```diff
@@ -1366,26 +1368,36 @@ class into the Health Check. Simply set the `predictorCategorization` to a
...
         from pdstools import __reports__
 
         verbose = kwargs.get("verbose", self.verbose)
 
-        if self.modelData is None or self.predictorData is None:
-            raise AssertionError("Needs both model and predictor data.")
+        if self.modelData is None:
+            raise AssertionError("Needs model data.")
+        if self.predictorData is None:
+            quarto = "HealthCheckModel.qmd"
```
StijnKas, line 1392: Please call this variable something like 'healthcheck_file' or something, not quarto. Quarto is a library, so let's not conflict

StijnKas (changes requested):

PR #98, opened 2023-05-30
Added formula for using Beta directly with positives

operdeck:
Added a note in both notebooks that we want to switch to using the positives directly.


operdeck:
Looks great

PR #99, opened 2023-05-30
Health Check consistency fixes

yusufuyanik1:
Health Check make up changes, some minor calculation fixes and expanding Model-only Health Check
- Treatment is added to hover of some plots if available
- Model data only Health Check expanded with additional plots
- weighed_average is used for overtime plots


python/pdstools/plots/plot_base.py
```diff
@@ -482,7 +482,7 @@ def plotOverTime(
                 df.groupby_dynamic("SnapshotTime", every=every, by=groupby)
                 .agg(
                     [
-                        (pl.sum("Positives") / pl.sum("ResponseCount")).alias(
+                        weighed_average_polars("SuccessRate", "ResponseCount").alias(
```
StijnKas, line 485: Why would we want to use weighed average success rate? I don't think that's right, is it?

StijnKas (changes requested):

PR #113, opened 2023-07-25
freeze polars version and article fixes

yusufuyanik1:


StijnKas (changes requested):
Thanks Yusuf, looks good. One thing: could you also update https://github.com/pegasystems/pega-datascientist-tools/blob/master/pyproject.toml? We unfortunately need to duplicate the requirements (or you could look into using just one source of truth for requirements if you'd like ;))

yusufuyanik1:
Tests look good after pyproject.toml update @StijnKas

PR #123, opened 2023-09-22
Made the off line reports shine even more

operdeck:
Fixed some text and made more plots interactive, consistent labels. Provided a better example of running the reports in batch.


operdeck:
Great stuff

PR #125, opened 2023-09-28
polars version upgrade

yusufuyanik1:


python/requirements.txt
```diff
@@ -3,7 +3,7 @@ plotly>=5.5.0
 requests
 pydot
 tqdm
-polars>=0.18.0,<=0.18.8
+polars==0.19.4
```
StijnKas, line 6: Let's be a bit more flexible: it's i think it works from `polars>=0.19.0, <=0.19.5`

PR #130, opened 2023-10-03
Dockerize HealthCheck

yusufuyanik1:
For testing:

- install docker from https://docs.docker.com/engine/install/
- go to the repo in the terminal
- build the image with this command: `docker build -t healthcheck .`
- Run the container with this command: `docker run -p 8501:8501 healthcheck`

Notes:

1. I am getting SSL verification errors for every website during build. Worked around it by just ignoring these checks. We can't productionize like that
2. Uploading zips works but CDHSample doesn't work as it needs internet access 
3. Investigate more efficient/faster base images/settings.


StijnKas:
I would like to hold off on merging this for now, I don't know what is the proper protocol here - whether to even include the dockerfile in github at all, upload artefacts, etc. Let's get that figured out soon.

PR #134, opened 2023-10-10
Issue 128

operdeck:
Standalone model report complete content-wise. Remains some TODOs to clean up code, move to PDS tools library and generally make sure it is robust and the numbers compare visavis the R version.

### DIFF
--- a/python/pdstools/reports/ModelReport.qmd
+++ b/python/pdstools/reports/ModelReport.qmd
@@ -26,7 +26,7 @@ import sys
 sys.path.append('/Users/perdo/Documents/pega/pega-datascientist-tools/python')
 
 from pdstools import datasets, ADMDatamart
-from pdstools.utils.cdh_utils import zRatio
+from pdstools.utils.cdh_utils import zRatio, lift
 import polars as pl
 from IPython.display import display, Markdown
 from tabulate import tabulate
@@ -77,9 +77,10 @@ display(
         f"""
 # {model_name_in_context}
 
-Channel: {channel_name}
+| Channel | ID |
+|:-----|------:|
+| {channel_name} | {model_id} |
 
-Internal ID: {model_id}
 """
     )
 )
@@ -101,8 +102,7 @@ fig.show()
 
 ::: {.callout-tip}
 The charts (built with [Plotly](https://plotly.com/python/)) have [user controls for panning,
-zooming etc](https://plotly.com/chart-studio-help/zoom-pan-hover-controls/) but
-note that these interactive plots do not render well in portals like Sharepoint
+zooming etc](https://plotly.com/chart-studio-help/zoom-pan-hover-controls/). These interactive plots do not render well in portals like Sharepoint
 or Box. It is preferable to view them from a browser.
 :::
 
@@ -123,14 +123,14 @@ auc_roc = round(classifier
 display(
     Markdown(
         f"""
-The model performance is {auc_roc} measured as AUC-ROC. This number is calculated from the “active” bins of the Classifier.
+The model performance is **{auc_roc}** measured as AUC-ROC. This number is calculated from the “active” bins of the Classifier.
 """
     )
 )
 
 ```
 
-The classifier is used to map the model scores (average of the log odds of the active predictors) to a propensity value. The “active” bins are the ones that can be reached from the current binning of the active predictors.
+The classifier maps the model scores (average of the log odds of the active predictors) to a propensity value. The “active” bins are the ones that can be reached from the current binning of the active predictors.
 
 See the [ADM Explained](https://pegasystems.github.io/pega-datascientist-tools/Python/articles/ADMExplained.html) article for more information on how ADM exactly works.
 
@@ -140,7 +140,8 @@ See the [ADM Explained](https://pegasystems.github.io/pega-datascientist-tools/P
 classifier_properties = classifier.select(
     pl.last("ResponseCount").alias("Responses"),
     pl.last("Positives"),
-    pl.format("{}%", (pl.last("Positives")*100/pl.last("ResponseCount")).round(3)).alias("Base Propensity")
+    pl.format("{}%", (pl.last("Positives")*100/pl.last("ResponseCount")).round(3)).alias("Base Propensity"),
+    pl.format("{}", pl.last("Performance").round(4)).alias("Performance")
 )
 
 show(classifier_properties.collect().to_pandas())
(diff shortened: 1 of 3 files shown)


python/pdstools/reports/ModelReport.qmd
```diff
@@ -16,26 +16,35 @@ format:
...
-        light: flatly
-        dark: darkly
+      light: flatly
+      dark: darkly
 jupyter: python3
 ---
 
 ```{python}
+import sys
+sys.path.append('/Users/perdo/Documents/pega/pega-datascientist-tools/python')
```
StijnKas, line 26: Don't forget to remove this at some point ;)

python/pdstools/reports/ModelReport.qmd
```diff
@@ -16,26 +16,35 @@ format:
...
 import os.path
+import plotly.express as px
 ```
 
 ```{python}
 #| tags: [parameters]
 
 # These parameters are overwritten when called externally
+
+# TODO see if we can align with the params of the Health Check
```
StijnKas, line 43: What do we need for this?

python/pdstools/reports/ModelReport.qmd
```diff
@@ -16,26 +16,35 @@ format:
...
+
 datafolder = os.path.expanduser("~/Downloads/tmp/")
```
StijnKas, line 45: Could you link this to a url location maybe?

python/pdstools/reports/ModelReport.qmd
```diff
@@ -44,29 +53,45 @@ predictordetails_activeonly = True
...
+datamart.predictorData = datamart.predictorData.filter(pl.col("ModelID") == model_id)
+
+# TODO ensure this is only one predictor snapshot, just in case? although it would be 
+# neat to show predictor evolution, if available
+```
+
+```{python}
+# TODO to make generic make sure these context keys really exist, do intersect with names in the model data
+
+channel_name = datamart.modelData.select(pl.concat_str(["Direction", "Channel"], separator="/")).unique().collect().item()
```
StijnKas, line 71: Can't you easily use your fancy `.format` function here?

python/pdstools/reports/ModelReport.qmd
```diff
@@ -139,21 +204,90 @@ The Cumulative Gains chart shows the percentage of he overall cases in the "posi
...
+    type='line', line=dict(dash='dash'),
+    x0=0, x1=100, y0=0, y1=100
+)
+fig.update_yaxes(scaleanchor="x", scaleratio=1, constrain='domain', title=dict(text="% of Positive Responders"))
+fig.update_xaxes(constrain='domain', title=dict(text="% of Population"))
+fig.show()
+```
+
+```{python}
+# TODO perhaps this should move into the pdstools plot functions "plotCumulativeLift"
```
StijnKas, line 228: Let's do it

python/pdstools/reports/ModelReport.qmd
```diff
@@ -139,21 +204,90 @@ The Cumulative Gains chart shows the percentage of he overall cases in the "posi
...
+
+:::
 
 # Trend charts
 
-TODO using model datamart show trend of auc, responses, ctr, maybe all in
-one with tabs
+TODO see if we can put the various trend charts in tabs - although this seems to require go not express, then gain .data[0] probably gives the go objects
+
+But maybe .data[0] gives the trace
```
StijnKas, line 245: yes, you can extract the go object from a px chart. I think it's under either .data[0] or .layout[0]

python/pdstools/reports/ModelReport.qmd
```diff
@@ -139,21 +204,90 @@ The Cumulative Gains chart shows the percentage of he overall cases in the "posi
...
+
+
+```{python}
+fig = datamart.plotOverTime('weighted_performance')
+
+fig.update_layout(title="Model Performance Trend", yaxis_title="ROC-AUC", xaxis_title = "")
+fig.update_layout(showlegend=False)
```
StijnKas, line 252: why have two separate arguments for this?

python/pdstools/reports/ModelReport.qmd
```diff
@@ -139,21 +204,90 @@ The Cumulative Gains chart shows the percentage of he overall cases in the "posi
...
+
+fig.show()
+```
+
+```{python}
+fig = datamart.plotOverTime('SuccessRate')
+
+fig.update_layout(title="Success Rate Trend", yaxis_title="Success Rate", xaxis_title = "")
+fig.update_layout(showlegend=False)
```
StijnKas, line 261: Same as above - I'd personally put all `update_layout` arguments in one call, easier to change after

python/pdstools/reports/ModelReport.qmd
```diff
@@ -139,21 +204,90 @@ The Cumulative Gains chart shows the percentage of he overall cases in the "posi
...
 
 Showing the performance across all predictors. The predictor categories default
 to the text before the first dot. This can be customized when reading the data
 for a particular customer.
 
-TODO add plot of overall predictor category performance
+```{python}
+#| error: true
+
+# TODO facets are meaningless but plotPredictorCategoryPerformance fails when
```
StijnKas, line 277: Should we fix this? Maybe there should be a default argument then

python/pdstools/reports/ModelReport.qmd
```diff
@@ -197,65 +328,135 @@ predictors. This can be configured via a parameter to this report.
...
+
+    # Add line for base propensity, TODO consider putting back in the library
+    fig.add_hline(y=base_propensity, line_dash="dash", line_color="grey", yref='y2')
+    # fig.update_xaxes(type='category') # prevent plotly from trying to guess
+
+    fig.update_layout(width=700, height=300)
+    fig.show()
+
+    # "Philip Mann" plot with simple red/green lift bars relative to base propensity
+    # TODO move below "Philip Mann" plot into a plot function (plotBinningLift)
```
StijnKas, line 388: Highlighting as not to forget

python/pdstools/reports/ModelReport.qmd
```diff
@@ -197,65 +328,135 @@ predictors. This can be configured via a parameter to this report.
...
+        template='pega'
+    )
+    fig.update_yaxes(type='category')
+    fig.layout.xaxis.tickformat = ",.2%"
+    fig.add_vline(x=0, line_color="black")
+    fig.update_layout(
+        yaxis = dict(autorange="reversed"),
+        title="Propensity Lift", xaxis_title="", yaxis_title="")
+    # Not convinced why we need below. It should respect the binindex order but w/o this it doesnt always.
+    fig.update_yaxes(categoryorder='array', categoryarray=pm_plot_binning_table['BinSymbol'])
```
StijnKas, line 425: I _think_, if there's no missing bin, it tries to cast to numeric and tries to do magic with sorting. If there's a missing bin, then casting to numeric will fail and it will just keep the order

python/pdstools/reports/ModelReport.qmd
```diff
@@ -197,65 +328,135 @@ predictors. This can be configured via a parameter to this report.
...
+    # TODO: when there are many bins not all are always shown, figure out how to fix
```
StijnKas, line 426: `nbins` argument? https://plotly.com/python/histograms/#choosing-the-number-of-bins

python/pdstools/reports/ModelReport.qmd
```diff
@@ -197,65 +328,135 @@ predictors. This can be configured via a parameter to this report.
...
+            pl.col("BinPositives").alias("Positives"),
+            zRatio(pl.col("BinPositives"), pl.col("BinNegatives")).round(3).alias("Z Ratio"),
+            (lift(pl.col("BinPositives"), pl.col("BinNegatives")) * 100).round(2).alias("Lift (%)"),
+            (pl.col("BinPropensity") * 100).round(3).alias("Propensity (%)")
+        )
+        .collect()
+    )
+
+    display(Markdown(
+        tabulate([[f[i] for f in human_friendly_binning_table] for i in range(human_friendly_binning_table.shape[0])],
```
StijnKas, line 448: Why this odd comprehension? Can't you just do `[i for i in range(human_friendly_binning_table)]`?

python/pdstools/utils/cdh_utils.py
```diff
@@ -615,6 +615,35 @@ def zRatioimpl(
...
+    Examples
+    --------
+    >>> df.group_by(['ModelID', 'PredictorName']).agg([lift()]).explode()
+    """
+
+    def liftImpl(binPos, binNeg, totalPos, totalNeg):
+        return (
+            # TODO not sure how polars (mis)behaves when there are no positives at all
+            # I would hope for a NaN but base python doesn't do that. Polars perhaps.
+            binPos * (totalPos + totalNeg) / ((binPos + binNeg) * totalPos)
```
StijnKas, line 643: It does have proper None value support, may work like you say

StijnKas (changes requested):

StijnKas:
Process Mining is going to get some abbreviation competition with Mr. Mann if he keeps thinking of new plots ;)

python/pdstools/reports/ModelReport.qmd
operdeck: to support calling from CLI - and HC quarto updated with the same now

python/pdstools/reports/ModelReport.qmd
operdeck: :)

python/pdstools/reports/ModelReport.qmd
operdeck: ?

StijnKas: I meant rather than having a hardcoded url to /downloads/tmp, are the files in github? Then you can just link to the file in github directly

python/pdstools/reports/ModelReport.qmd
operdeck: I need a transposed list of lists

python/pdstools/reports/ModelReport.qmd
operdeck: The plot functions are a little daunting, I'll first get the quarto feature complete then move things around

python/pdstools/reports/ModelReport.qmd
operdeck: See below

python/pdstools/reports/ModelReport.qmd
operdeck: Maybe you can build a test that fails if such lines are still there. I do something similar for the R version, a test fails if it finds un-commented library source inclusion.

StijnKas: Right yeah I think something like that would be cool. Probably in tandem with adding default linters/formatters. I'll create an issue for that, wanted tot do it anyway

operdeck:
Will send a new one in - most remarks left as TODOs for now

PR #135, opened 2023-10-10
Add devcontainer in support for codespaces

StijnKas:
This one supersedes #130 - adding devcontainer with proper configuration makes 1) codespaces work, 2) docker creation seamless and 3) environments consistent.

Building the environment takes quite a while, because we don't just install a pretty sizable python image but also install all app dependencies & quarto. Let's see if we can improve this after.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -16,7 +16,10 @@
     "portsAttributes": {
         "8501": {
             "label": "Pdstools App",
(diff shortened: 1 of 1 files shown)


StijnKas:
@operdeck @yusufuyanik1 could you give the new codespaces thing a shot? Should be as simple as just hitting the `code` button on the homepage and then clicking `Create codespace on master`. Be sure to kill your codespace after using it on https://github.com/codespaces

PR #144, opened 2023-10-17
Python 3.12 support, bump version

StijnKas:
This PR supports Python 3.12. 

On our side, we should be good to go for 3.12, but some of our dependencies don't support it yet.

This PR bumps the tests to include 3.12 - once those pass, we can officially state 3.12 support.


StijnKas:
Currently failing because PyArrow does not support python 3.12 yet. See https://github.com/apache/arrow/issues/37880. Looks like we need to wait for Arrow 14.0.0, which is scheduled in about a week.

StijnKas:
Update: pyarrow issue is resolved, now waiting for aiohttp's 3.9 release.

PR #171, opened 2023-12-04
HealthCheck fix

yusufuyanik1:
After this change,
If context keys provided by the user(if not provided, it is set to the default ["Channel", "Direction", "Issue", "Group"]) are not present in dataframe we add these columns ourselves and fill them with "NA". 
For example, when we run HC on a datamart without "Channel" and "Direction" columns. If the user leaves this section as is:

Screenshot 2023-12-04 at 16 20 35

it should run without any problem, showing NA value for the Channel and Direction. 

Hovewer, if the user removes for example **Channel** from the list above, since "Channel" is hardcoded in the HealthCheck.qmd, we will have an error. Making the script more robust by removing hard coded column names makes sense but it adds some if-else's, making the script more crowded

Lets discuss here if there is a better solution. @operdeck @StijnKas


operdeck:
Agree on the approach. I think that whatever we distill from pyName should
just be added, keeping it simple. Changing context used to be more common
than it is now but still, we have big accounts doing this. And in Process
AI use cases it happens a lot.

We need to make our plots more robust against variations in context keys.


Op ma 4 dec 2023 om 18:07 schreef Stijn Kas ***@***.***>

yusufuyanik1:
In that case, we will need to revise the HealthCheck.qmd file to replace hardcoded context keys. Following the approach described by Stijn, an error will occur if the Channel column is not present in the data. I will update once it is done

Issue #191, opened 2024-01-22
Github Codespaces File Uploading


yusufuyanik1:
### pdstools version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/pdstools/) of pdstools.

### Issue description

In Github Codespaces streamlit.file_uploader fails to upload data bigger than 50mb. 

### Reproducible example

```Python
Can't reproduce.
```

### Expected behavior

Files up to 2 gb to be <img width="740" alt="Screenshot 2024-01-22 at 14 07 48" src="https://github.com/pegasystems/pega-datascientist-tools/assets/54443450/2d9367e0-8609-47d0-9373-5d24adac82b4"> handled

### Installed versions

<details>

```
---Version info---
pdstools: 3.3.0
Platform: macOS-13.6.3-arm64-arm-64bit
Python: 3.11.5 (main, Sep 11 2023, 08:31:25) [Clang 14.0.6 ]

---Dependencies---
plotly: 5.18.0
requests: 2.31.0
pydot: 1.4.2
polars: 0.20.2
pyarrow: 14.0.2
tqdm: 4.66.1
pyyaml: <not installed>
aioboto3: 12.1.0

---Streamlit app dependencies---
streamlit: 1.29.0
quarto: 0.1.0
papermill: 2.5.0
itables: 1.6.3
pandas: 2.1.4
jinja2: 3.1.2
xlsxwriter: 3.1.9```

</details>

PR #192, opened 2024-01-22
Guide user in CodeSpaces to upload file

yusufuyanik1:
How about this solution for this [issue](https://github.com/pegasystems/pega-datascientist-tools/issues/191#issue-2093895543) about file_uploader problem?

If the working directory is /workspaces/pega-datascientist-tools then we hide the file_uploader option and pre-populate direct file path option with the path.

@StijnKas @operdeck

### DIFF
--- a/python/pdstools/utils/streamlit_utils.py
+++ b/python/pdstools/utils/streamlit_utils.py
@@ -28,18 +28,16 @@ def cachedDatamart(*args, **kwargs):
 def import_datamart(**opts):
     st.session_state["params"] = {}
     st.write("### Data import")
-    options = [
-        "Direct file path",
-        "Direct file upload",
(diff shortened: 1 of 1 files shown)


StijnKas:
@yusufuyanik1  conceptually i'm sort-of fine with that but we need to give a clear explanation/instruction on where exactly to put the data whenever we hit this. 

What i may prefer is to give a warning (`st.warning`) whenever we detect that as a wd, still give the option to upload files, but use `/workspaces/pega-datascientist-tools` as the default value for the filepath upload. That way we don't _remove_ functionality in codespaces, we just detect when you may run into issues and help you along. In some cases (small data) the file upload may still continue to work, just more effort for us to maintain a version where the upload buttons are actually removed.

PR #208, opened 2024-04-30
Revert incorrect depcrecation

StijnKas:
Fyi @yusufuyanik1  - I think you accidentally edited this to avoid a polars rename, but this was python code rather than Polars. No harm, but I've changed it back here.


yusufuyanik1 (approved):
Correct, my mistake

PR #233, opened 2024-06-18
[WIP] Compatibility with Polars V1

StijnKas:
Polars V1 brought quite some breaking changes, this PR gets us ready for when it releases, and would tie our Polars version compatibility to Polars V1.0.x. Given how core Polars is to the functionality of pdstools, we can't _always_ use the latest version, but given that we'd like to stay up to date, we'll tie ourselves to each minor version. Once a new minor version becomes available, we'll retest our changes on the increment and see if we can relieve this requirement.

As part of this PR, the main changes I had to cover were:
- `StringCache()` no longer took True as an argument
- The `strict` argument on DataFrame/LazyFrame/Series became more .. strict, so I turned it off in the short-term (this was not critical anywhere anyway, moreso within the testing code)
- Group_by and Partition_by keys are now tuples instead of strings when partitioning/grouping by only one key. Needed to unpack them
- Added a missing test for `model_summary`
- `suffix`/`prefix` on `expr`s are now part of the `name` namespace

Also added some QOL features such as checking Pandoc version & adding TestBook as an optional dependency.


StijnKas:
@yusufuyanik1  @operdeck  could you help me validate this branch on your workflows and fix any issues you may come across? We should probably get to polars V1 sooner rather than later, given we're only going to fall further behind and run into more issues down the line

StijnKas:
@yusufuyanik1 @operdeck Could both of you look at this tomorrow? I will too, let's get it merged so we are semi-up to date again

StijnKas:
FYI @yusufuyanik1 , we're also getting a ton of warnings in the health check execution: even things like 'where' being used over 'filter', which I think is a relatively old change. Let's get those out of the way as well

Issue #205, opened 2024-04-02
Glob method for combining multiple zip files in readMultizip in Pega_IO


brianhamnerPega:
### Is your feature request related to a problem?

No

### Describe the solution you'd like

When exporting data mart data from Pega, we need to use data flows (due to security settings in browsers that block file downloads). This often results in many different zip files. To prep this for PDSTools, we need to merge all the zip (really the json within) to a single file so that ADMDatamart class can operate on it. 

In the current file_io there is a readMultiZip function that reads and combines multiple zips. However you need to handle specifying the glob pattern eg:

path = "/Users/hamnb/Downloads/ADM2/"

pattern="Data-Decision-ADM-ModelSnapshot_pzModelSnapshots*"
model_files = glob.glob(os.path.join(path, pattern))

pattern = "Data-Decision-ADM-PredictorBinningSnapshot*"
bin_files = glob.glob(os.path.join(path, pattern))

Once you have these lists of files, you can call the readMultiZip to get the combined file. 

Would be nice to have this all available and documented, in both the notebook and quarto examples, where you simply supply the path, and whether it is a mutli-zip file (or multi json).

### Describe alternatives you've considered

Wrote the glob pattern in our notebook.

PR #235, opened 2024-06-18
Updated and heavily simplified anonymization script

StijnKas:
We've had an anonymization script in the tools for a little bit, but these were not performant enough on any realistic and real loads, so it was time for an update. The configuration options here are much less, but it's much more efficient.

It utilizes a two-pass approach, whereby we first output all files to batched parquet files and then loop over all parquet files to generate one single output parquet file. 

Many thanks to  @danielm-dk for helping improve this part.


python/pdstools/pega_io/Anonymization.py
```diff
@@ -0,0 +1,314 @@
...
+        temp_path (str):
+            Path to the temporary directory where the Parquet file will be saved.
+        i:
+            Index of the chunk.
+
+        Returns:
+        str: File path of the converted Parquet file.
+        """
+
+        init_df = pl.concat([pl.read_ndjson(n) for n in files], how="diagonal_relaxed")
```
yusufuyanik1, line 194: can we use readDSExport in order to be able to accommodate other data formats(parquet, arrow, csv...) too?

StijnKas: Good point. Right now I don't think we can easily because we're doing batching based on a glob pattern here, which I don't think readDSExport supports. This is a feature request we've received in the past though (#205), so once we've got this functionality in, we should certainly use it here as well!

StijnKas:
@yusufuyanik1  would you mind giving this a review? Would like to merge it sometime soon

yusufuyanik1 (approved):
Looks good, worked with some sample data i created :)

PR #242, opened 2024-06-26
fix safeName function

yusufuyanik1:
There was a bug with the safeName function when we had both json and string in pyName column. When pyName column looks like this:

1. '{"pyName":"CreditCard","pyTreatment":"Gold"}'
2. 'Iphone'

`df.select(safeName().str.json_decode()).unnest("tempName")`
This returns only pyName  I assume because of this conflict where the pyTreatment column has both a string from first row ("Gold") and a None type from second row. 

Now i am filling pyTreatment with an empty string if it does not exist and this fixes the issue.

@StijnKas please review this.
@operdeck fyi this is why we didn't get the pyTreatment column.

### DIFF
--- a/python/pdstools/utils/cdh_utils.py
+++ b/python/pdstools/utils/cdh_utils.py
@@ -6,17 +6,18 @@ Various utilities to access and manipulate data from Pega for purposes
 of data analysis, reporting and monitoring.
 """
 
-from typing import List, Union
-import polars as pl
-import re
-import numpy as np
 import datetime
+import re
 import warnings
-from .types import any_frame
+from typing import List, Union
+
+import numpy as np
+import polars as pl
+import pytz
+
 from .errors import NotEagerError
 from .table_definitions import PegaDefaultTables
-
-import pytz
+from .types import any_frame
 
 
 def defaultPredictorCategorization(
@@ -86,7 +87,16 @@ def _extract_keys(
     def safeName():
         return (
             pl.when(~pl.col(col).cast(pl.Utf8).str.starts_with("{"))
-            .then(pl.concat_str([pl.lit('{"pyName":"'), pl.col(col), pl.lit('"}')]))
+            .then(
+                pl.concat_str(
+                    [
+                        pl.lit('{"pyName":"'),
+                        pl.col(col),
+                        pl.lit('",'),
+                        pl.lit('"pyTreatment":""}'),
+                    ]
+                )
+            )
             .otherwise(pl.col(col))
         ).alias("tempName")
 
@@ -576,6 +586,7 @@ def fromPRPCDateTime(
 
 # TODO: Polars doesn't like time zones like GMT+0200
 
+
 def toPRPCDateTime(dt: datetime.datetime) -> str:
     """Convert to a Pega date-time string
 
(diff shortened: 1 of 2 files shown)


operdeck:
This assumes pyName is only encapsulating Name + Treatment, which isn't true. You can configure anything in the model context and whatever is outside of the std context fields, gets respresented as JSON in pyName. We have some customers including e.g. a category in there.

StijnKas:
> This assumes pyName is only encapsulating Name + Treatment, which isn't true. You can configure anything in the model context and whatever is outside of the std context fields, gets respresented as JSON in pyName. We have some customers including e.g. a category in there.

Agreed. Maybe it would be useful to have some minimum reproducible example here so we can see exactly why it was failing before, because it's not entirely clear to me and it may help us find a more general solution?

operdeck:
The alt approach here is to take the unique pyNames, then only un-json the ones that start with a { into a new frame that will have all the new columns. Then merge back that new frame into the existing df. Might be faster as well. I did something along those lines in the R version. Merging back is a little tricky because columns may already exist and should be overwritten etc.

StijnKas:
@yusufuyanik1  is this still relevant? If so, looks like it needs a rebase. But given by the above discussion, feels like there's still some work to be done?

yusufuyanik1:
`json_decode` was the problem, the default `infer_schema_length` is 100 and if there are no treatments (only pyName) in the first 100 rows, then it is not parsing other columns in pyName.

Below is a reproducible example, in the first 100 rows of pyName we have string and then comes the json.
If the infer_schema_length of json_decode is not set, it won't parse pyTreatment
 ```
import polars as pl
df = pl.DataFrame(
    {
        "pyName": ["GoldCard"] * 100
        + ['{"pyName":"SilverCard","pyTreatment":"Silver_Web"}']
    }
)

col = "pyName"
def safe_name():
        return (
            pl.when(~pl.col(col).cast(pl.Utf8).str.starts_with("{"))
            .then(pl.concat_str([pl.lit('{"pyName":"'), pl.col(col), pl.lit('"}')]))
            .otherwise(pl.col(col))
        ).alias("tempName")


series = (
        df.select(
            safe_name().str.json_decode(),
        )
        .unnest("tempName")
        .lazy()
        .collect()
    )
series
```
Screenshot 2024-08-16 at 11 56 33
Should we set this to None? @StijnKas

operdeck:
Sort first so any { is present in the first 100?

Group by configuration name and check for each of those separately then do
an any?


Op vr 16 aug 2024 om 11:57 schreef yusufuyanik1 ***@***.***>

operdeck:
Maybe take just head(1) per configuration - I think either all or none of
the pyName fields are json , per model rule / configuration. Unless someone
changes that over time :( So perhaps by config + timestamp?

Op vr 16 aug 2024 om 14:33 schreef yusufuyanik1 ***@***.***>

yusufuyanik1:
infer_schema_lenght was set to None in this [commit](https://github.com/pegasystems/pega-datascientist-tools/commit/1cde6497be64587fbba5123023501b02896b130b#diff-7cac4e15cab78a4ad078fe4cea8553d5599ed430af3101b3e16497bfc78e9bd0R128)

This operation is being applied to modelData, which is a moderately sized(in terms of row count). Also, we are already collecting this column later 2 rows below, so i don't think it is a big deal.

Issue #213, opened 2024-05-15
_delete_temp_files in generateReport is _way_ too aggressive.


StijnKas:
### pdstools version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/pdstools/) of pdstools.

### Issue description

The `_delete_temp_files` method in generateReport of the ADMDatamart class is _way_ too aggressive.

https://github.com/pegasystems/pega-datascientist-tools/blob/master/python/pdstools/adm/ADMDatamart.py#L1657

By default, it currently deletes _all_ [".yaml", ".qmd", ".ipynb", ".arrow"] files in the working directory. 

Considering that, by default, the working directory is the current directory, this will remove all files with these types from the local directory, and since it's using OS commands they're not put in the thrash either. It already accidentally removed a few notebooks and other cached files that i had there locally. 

### Reproducible example

```Python
Have files in your current directory, then run generateReport
```

### Expected behavior

We should make sure it only deletes the cached files that we write for the health check, and only the files used to generate this specific health check run. 

### Installed versions

<details>

```
Replace this line with the output of pdstools.show_versions(), leave the backticks in place
```

</details>

StijnKas:
Closed by https://github.com/pegasystems/pega-datascientist-tools/pull/253

PR #250, opened 2024-07-23
Windows error

yusufuyanik1:
- Added warnings about Pandoc and pdstools version to Health Check app.
- Seperated generateReport function to generate_health_check and generate_model_reports
- Changed the temp_dir logic. Everything is written to a freshly created temp_dir and the whole folder is removed if we are not in debug mode.

### DIFF
--- a/python/pdstools/adm/ADMDatamart.py
+++ b/python/pdstools/adm/ADMDatamart.py
@@ -226,7 +226,7 @@ class ADMDatamart(Plots, Tables):
         If model_df or predictor_df is supplied, it will use those instead
         If any filters are included in the the `query` argument of the ADMDatmart,
         those will be applied to the modeldata, and the predictordata will be
-        filtered such that it only contains the modelids leftover after filtering.
+        filtered such that it only contains the model_ids leftover after filtering.
         After reading, some additional values (such as success rate) are
         automatically computed.
         Lastly, if there are missing columns from both datasets,
@@ -1664,10 +1664,10 @@ Meaning in total, {self.model_stats['models_n_nonperforming']} ({round(self.mode
     def generate_model_reports(
         self,
         name: Optional[str] = None,
-        model_list: List[str] = [],
+        model_list: List[str] = None,
         working_dir: Optional[Path] = None,
+        only_active_predictors: bool = False,
         *,
-        predictordetails_activeonly: bool = False,
         base_file_name: str = None,
         output_type: str = "html",
         debug_mode: bool = False,
@@ -1685,7 +1685,7 @@ Meaning in total, {self.model_stats['models_n_nonperforming']} ({round(self.mode
             The list of model IDs to generate reports for.
         working_dir : Path, optional
             The working directory for the output. If None, uses current working directory.
-        predictordetails_activeonly : bool, default=False
+        only_active_predictors : bool, default=False
             Whether to only include active predictor details.
         base_file_name : str, optional
             The base file name for the generated reports. Defaults to None.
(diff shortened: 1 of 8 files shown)


python/pdstools/adm/ADMDatamart.py
```diff
@@ -1654,174 +1661,325 @@ def overall_summary(
...
-        verbose = kwargs.get("verbose", self.verbose)
+        Parameters
+        ----------
+        name : str, optional
+            The name of the report.
+        working_dir : Path, optional
+            The working directory for the output. If None, uses current working directory.
+        output_type : str, default='html'
+            The type of the output file (e.g., "html", "pdf").
+        debug_mode : bool, optional
```
StijnKas, line 1793: Debug mode doesn't sound like the proper argument name for this description, wouldn't it be something like `delete_temp_dir`? debug mode sounds like you'd be logging more information while running

StijnKas (changes requested):
Some small comments, but thanks for this @yusufuyanik1, looks great overall!

StijnKas:
Resolves [#213](https://github.com/pegasystems/pega-datascientist-tools/issues/213)

python/pdstools/adm/ADMDatamart.py
yusufuyanik1: We already have the verbose argument for logging
1. Use verbose for both logging and keeping temp_dir
2. Create a new argument, debug_mode which overwrites verbose
`debug_mode = kwargs.get("debug_mode", self.verbose)`
So when it is set to True, it keeps logs and doesn't delete temp_dir

yusufuyanik1: Thanks for the notes @StijnKas, now all your notes are addressed except this. Do you think this is fine or should we separate the verbose and delete temp_dir logic?

StijnKas: Logically I don't think they're the same thing but I follow your logic and I think it's fine to make it easier on the user. Let's stick with this.

python/pdstools/app/health_check/Home.py
```diff
@@ -1,9 +1,9 @@
 import shutil
 
-import requests
 import streamlit as st
 
 import pdstools
+from pdstools.utils.streamlit_utils import get_latest_pdstools_version
```
StijnKas, line 6: Last remark: I don't think this should be in streamlit_utils, it's not streamlit specific

python/tests/test_healthcheck.py
```diff
@@ -103,9 +105,9 @@ def test_GenerateModelReport(sample):
 def test_GenerateModelReport_Failing(sample_without_predictorbinning):
     with pytest.raises(Exception) as e_info:
         sample_without_predictorbinning.generate_model_reports(
-            name="MyOrg", modelid=["bd70a915-697a-5d43-ab2c-53b0557c85a0"]
+            name="MyOrg", model_list="bd70a915-697a-5d43-ab2c-53b0557c85a0"
```
StijnKas, line 108: Well now you've made it model_list and are passing a string :)

python/pdstools/app/health_check/Home.py
StijnKas: Ah I see, you want it to be cached. Maybe it could be better to use some other caching mechanism so it's streamlit agnostic?

python/tests/test_healthcheck.py
yusufuyanik1: It is to make the test fail as the name suggests 😄

StijnKas: well look at me not reading :D

python/pdstools/app/health_check/Home.py
yusufuyanik1: Right, I used streamlit's caching decorator, will try to use another way of caching

yusufuyanik1: Carried the original function to cdh_utils and calling this from streamlit_utils now.
```
@st.cache_data
def st_get_latest_pdstools_version():
    return cdh_utils.get_latest_pdstools_version()

```

StijnKas (approved):

PR #253, opened 2024-08-22
Generate Report Refactor

yusufuyanik1:
Refactor health check and model report generation functions

- Reworked generate_health_check and generate_model_report functions
- Removed debug mode and verbose parameters 
- keep_temp_files parameter added to both functions
- Improved temporary file handling:
  - New temporary folder created for each run
  - Folder automatically deleted after report generation or on error, unless keep_temp_files is True
  - Eliminated risk of deleting unrelated files
- logger is set on the pdstools app home page and it is persisted into ADMDatamart. A log file with timestamp is created if we get an error while generating report in the app
- Temporarily fixed numpy version to <2
  - Will update to the latest version in the V1 branch
  - Updates for polars and streamlit also planned for V1 branch


python/pdstools/app/health_check/pages/3_Reports.py
```diff
@@ -101,22 +99,27 @@
...
-                data=f,
-                file_name="errorlog.txt",
-                key="ErrorLogDownload",
-            )
+        if "HCErrorLogDownload" not in st.session_state:
+            st.error(f"""An error occurred hc: {e}""")
+            traceback_str = traceback.format_exc()
+
+            # Find the most recent log file in the working directorpdsy
+            log_files = list(working_dir.glob("pdstools_logs_*.txt"))
```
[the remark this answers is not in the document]
yusufuyanik1: Shall we keep this user flexibility to set working_dir? I don't see how this can be dangerous.
Screenshot 2024-08-22 at 17 59 11

python/pdstools/adm/ADMDatamart.py
```diff
@@ -152,6 +153,8 @@ def __init__(
         verbose: bool = False,
         **reading_opts,
     ):
+        self.logger = logging.getLogger(__name__)
```
StijnKas, line 156: I don't think you should set a logger to an object? Isn't the convention to set a logger at file level (i.e., just set `logger=logging.getLogger(__name__)` and simply do `logger.info()`?

python/pdstools/adm/ADMDatamart.py
```diff
@@ -152,6 +153,8 @@ def __init__(
...
+        self.logger.setLevel(logging.DEBUG)
```
StijnKas, line 157: Same with above

python/pdstools/adm/ADMDatamart.py
```diff
@@ -1719,18 +1725,16 @@ def generate_model_reports(
...
         working_dir.mkdir(parents=True, exist_ok=True)
-
-        # Create a unique temporary directory name
         temp_dir_name = (
             tempfile.mkdtemp(prefix=f"tmp_{name}_", dir=working_dir)
             if name
             else tempfile.mkdtemp(prefix="tmp_", dir=working_dir)
         )
         temp_dir_path = Path(temp_dir_name)
+        timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
```
StijnKas, line 1736: I would suggest to use the ISO standard (https://www.iso.org/iso-8601-date-and-time-format.html). Main change is to remove the `_` and add milliseconds.

python/pdstools/adm/ADMDatamart.py
```diff
@@ -1811,18 +1823,17 @@ def generate_health_check(
...
```
StijnKas, line 1835: Why the duplication here?

python/pdstools/adm/ADMDatamart.py
```diff
@@ -1898,13 +1913,26 @@ def _write_params_file(self, temp_dir, model_id, only_active_predictors):
...
+        log_file: Path,
+        temp_dir: Path,
+        qmd_file: str,
+        output_type: str,
+        output_filename: str,
+        save_log_file: bool,
+    ) -> int:
         """Run the Quarto command to generate the report."""
+
+        def log_and_write(message: str, level: str = "INFO"):
```
StijnKas, line 1926: Honestly I would try to use OOTB logging as much as you can, simply saying `logger.info()`. You can then intercept your logger in the streamlit app or wherever else to capture/intercept and write to a file?

StijnKas: Don't want to be a pain here, but the more custom the `log_and_write` function becomes, the further away it becomes from 'recognizable' python code. That's something I want to watch out for.

python/pdstools/adm/ADMDatamart.py
```diff
@@ -1916,10 +1944,9 @@ def _run_quarto_command(
                 check=True,
             )
             quarto_version = version_result.stdout.strip()
-            logging.info(f"Quarto version: {quarto_version}")
-            # add version compatibility check here if needed
+            log_and_write(f"Quarto version: {quarto_version}")
         except subprocess.CalledProcessError as e:
-            logging.warning(f"Failed to check Quarto version: {e}")
+            log_and_write(f"WARNING: Failed to check Quarto version: {e}")
```
StijnKas, line 1949: Just one of those examples where I think OOTB logging makes more sense: `logging.warning()` makes your implicit warning here, explicit.

python/pdstools/app/health_check/pages/3_Reports.py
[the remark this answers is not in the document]
StijnKas: I agree - but I suppose since you're calling a glob on top of a working_dir variable, it may lead to variance. Maybe it would be more reliable to put the working_dir in an f-string?

python/pdstools/plots/plot_base.py
```diff
@@ -672,12 +672,13 @@ def plotScoreDistribution(
             "BinPropensity",
             "ModelID",
         }
+
         df, _ = self._subset_data(table, required_columns, query)
         if modelids is not None:
             df = df.filter(pl.col("ModelID").is_in(modelids))
-        df = df.filter(pl.col("PredictorName") == "Classifier").collect()
+        df = df.filter(pl.col("PredictorName") != "Classifier").collect()
```
StijnKas, line 679: Why flip this sign? I'm not sure if that looks right?

yusufuyanik1: Nice catch, that was a mistake, thought it was a different function.

python/pdstools/adm/ADMDatamart.py
StijnKas: As far as I know, logger handlers are at an application level, not a script level. So the convention is to simply use OOTB loggers on the script level (namely, the getLogger(__name__) object), so that consumers can decide how to read these logs. 

Streamlit is one of those application/consumers, so in the streamlit app itself we could then create such a logging stream handler to capture logs and put into a text file. Therefore, I wouldn't expect this code in the ADMDatamart.py file? Perhaps I'm misunderstanding though, so please feel free to correct me.

yusufuyanik1:
Thanks for the review @StijnKas. I believe I've addressed your comments. Could you please review them again when you have time?

python/pdstools/adm/ADMDatamart.py
```diff
@@ -24,6 +22,9 @@
 from .ADMTrees import ADMTrees
 from .Tables import Tables
 
+logger = logging.getLogger("PDSTools.ADMDatamart")
+logging.basicConfig(level=logging.INFO)
```
StijnKas, line 26: I don't think you need this

python/pdstools/adm/ADMDatamart.py
```diff
@@ -24,6 +22,9 @@
 from .ADMTrees import ADMTrees
 from .Tables import Tables
 
+logger = logging.getLogger("PDSTools.ADMDatamart")
```
StijnKas, line 25: I would simply do `logging.getLogger(__name__)`. No need to customize - `__name__` would already produce a unique path

python/pdstools/adm/ADMDatamart.py
```diff
@@ -1914,25 +1885,17 @@ def _write_params_file(self, temp_dir, model_id, only_active_predictors):
...
-            getattr(self.logger, level.lower())(message)
-            if save_log_file:
-                with open(log_file, "a") as log:
-                    log.write(f"{message}\n")
-
         try:
             quarto_exec = self._find_quarto_executable()
         except FileNotFoundError as e:
-            log_and_write(f"ERROR: {str(e)}")
+            logger.warning(e)
```
StijnKas, line 1898: Should this not be logger.exception?

python/pdstools/adm/ADMDatamart.py
```diff
@@ -1944,9 +1907,9 @@ def log_and_write(message: str, level: str = "INFO"):
                 check=True,
             )
             quarto_version = version_result.stdout.strip()
-            log_and_write(f"Quarto version: {quarto_version}")
+            logger.info(f"Quarto version: {quarto_version}")
         except subprocess.CalledProcessError as e:
-            log_and_write(f"WARNING: Failed to check Quarto version: {e}")
+            logger.warning(f"Failed to check Quarto version: {e}")
```
StijnKas, line 1912: Same as above

python/pdstools/utils/cdh_utils.py
```diff
@@ -1015,3 +1018,33 @@ def get_latest_pdstools_version():
         return response.json()["info"]["version"]
     except:
         return None
+
+
+def setup_logger():
+    logger = logging.getLogger("PDSTools")
```
StijnKas, line 1024: Would recommend to just use 'pdstools' lowercase, as that's the internal lib name

StijnKas (commented):
Hey, reviewed your further changes - looking better! Could you go through and resolve the comments you resolved, then I'll pass through it one last time.

python/pdstools/adm/ADMDatamart.py
yusufuyanik1: I am logging the quarto outputs at INFO level
```{python}
process = subprocess.Popen(
            command,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            cwd=temp_dir,
            text=True,
        )

while True:
    output = process.stdout.readline()
    error = process.stderr.readline()
    if output:
        logger.info(output.strip())
    if error:
        logger.info(error.strip())
    if process.poll() is not None and not output and not error:
        break
```

python/pdstools/adm/ADMDatamart.py
yusufuyanik1: I am using a child of the logger created in Streamlit app which has a log_buffer. If I initialize the buffer with ```__name__``` in ADMDatamart I can't combine the logs from the app and ADMDatamart.

python/pdstools/adm/ADMDatamart.py
StijnKas: Sure - but you shouldn't need to use basicConfig configuration in the logging _writing_ scripts, only the consuming? To write as info logging, simply do `logger.info()`. Unless I'm missing something?

python/pdstools/adm/ADMDatamart.py
yusufuyanik1: Done

python/pdstools/adm/ADMDatamart.py
yusufuyanik1: Not sure if this has to be an error. Quarto version check might fail as long as the quarto process runs

python/pdstools/utils/cdh_utils.py
yusufuyanik1: ```logger = logging.getLogger()```
Not we are configuring the root level logger in setup_logger because
```pdstools.adm.ADMDatamart.py```
```pdstoosl.app.health_check.Home.py```

The app logger is configured in Home.py but it is not a parent of ADMDatamart in folder hierarchy. So, I am setting the root level logger in the app since it is the common parent.

StijnKas (approved):

PR #257, opened 2024-09-09
Make Quarto Render OS-Agnostic and Add Version Info to Health Check Logs

yusufuyanik1:
- Simplified Quarto render subprocess to fix Windows failures and make it OS-agnostic.
- Added version info to Health Check error logs for improved debugging.


StijnKas (approved):
No comments on this code itself: looks good! My only request is that we find a way to dynamically generate the dependencies and optional dependencies directly from the pyproject toml file, but it looks like that's not entirely trivial and probably deserves its own issue.

PR #260, opened 2024-09-17
Pdstools V4

StijnKas:
This PR contains the initial version of the pdstools V4 release. We will first release an alpha release of version 4, as this is a pretty big upgrade.

V4 brings some pretty major (and necessary) changes. A lot of them are, unfortunately, breaking - but it's for the best. pdstools is now much easier to maintain and keep consistent, and new functionality now has a much more logical place to go. 

My goal is for the initial V4 release to contain most of the breaking (API-centric) changes we foresee in a long time. Then, we can of course still change the inner functionality and/or add new functions - but hopefully the most important function schemas/API don't need more changes anytime soon.

## ✨Highlights 
- Farewell R - you've served us well, but PDSTools is now Python only
- Introducing the Pega DX API Client
    - Starting out with support for the 24.2 Prediction Studio and Knowledge Buddy APIs
- **Major** refactor of the entire codebase: consistent python naming, optional dependency groups, well-defined typehints

## ❌Deprecations/removals 
- The R version of PDS Tools has been removed. In case you still want to use the R tools, you should manually clone the repo at the V3.x tag.
- The legacy IH utilities have been dropped. These were old parts of the codebase and untested/unused. New IH utilities are on their way!

## 🔨Changes 
- Consistent pythonic casing, meaning `PascalCase` for classes & `snake_case` for methods, variables & arguments
- Much improved typehints, so it's much more obvious what the response of a given function will be
- Fewer 'base' dependencies; different functionality is split up into 'namespaces' that all have their own set of requirements
    - The first time you invoke a method in a 'namespace', it verifies the dependencies and gives a clear warning if any are missing
- To expand on the previous point: functionality is split up much more logically. Taking the ADMDatamart class as an example:
    -  Plotting functionality is part of `ADMDatamart.plot.bubble_chart()` (or any other plot of course)
    - The health check and other reports are part of `ADMDatamart.generate.health_check()` (for instance)
    - The intermediate aggregations needed are part of `ADMDatamart.aggregations.pivot()` (for instance)
- Using `classmethod`s, we can initialize the ADMDatamart class in particular in a much more flexible way. 
    - The main `__init__` method of the ADMDatamart class is very simple: it expects two `polars.LazyFrame`s; one for `model_data` and one for `prediction_data`. If you've already read in your data, simply use this
    - If, instead, you want to use the previous functionality which automatically found the most recent file in a folder, you should initialize the datamart class like `ADMDatamart.from_ds_export()`
    - Or, if instead, you are consuming the results of a data flow (including the OOTB Prediction Studio export), you can simply initialize the datamart class like `ADMDatamart.from_dataflow_export(model_data="pattern_for_model_files*.json", predictor_data="pattern_for_predictor_files*.json")`. We can also cache the files we've read in before by writing to a 'cache' file automatically - this makes things move quickly. This closes https://github.com/pegasystems/pega-datascientist-tools/issues/205 as well.


python/pdstools/__init__.py
```diff
@@ -10,22 +10,53 @@
...
-from .utils.datasets import CDHSample, SampleTrees, SampleValueFinder
-from .utils.polars_ext import *
-from .utils.show_versions import show_versions
-from .utils.table_definitions import PegaDefaultTables
-from .valuefinder.ValueFinder import ValueFinder
+
+# from .utils.polars_ext import *
+# from .utils.show_versions import show_versions
+# from .valuefinder.ValueFinder import ValueFinder
+from .infinity import Infinity
```
yusufuyanik1, line 17: I need httpx when importing pdstools because it imports Infinity which requires httpx. @StijnKas

StijnKas: Good spot @yusufuyanik1 . Need to make sure https stays optional :)

python/pdstools/reports/ModelReport.qmd
```diff
@@ -103,22 +103,23 @@ kwargs = dict()
 ```{python}
 # Predictor data for one model ID
```
operdeck, line 104: @yusufuyanik1  perhaps we need to fix something here to support the call from streamlit, pls have a look

python/pdstools/__init__.py
```diff
@@ -44,7 +45,7 @@
     "read_ds_export",
     "setupAzureOpenAI",
     "Prediction",
-    "NBAD",
```
operdeck: @StijnKas Need help to sort out the change from NBAD to CDH_Guidelines. test_Prediction only fails because of some of this dependency stuff I think.

StijnKas: CDH_Guidelines is now in the `adm` folder - but I'm not entirely sure what you're trying to access in your Predictions code, @operdeck ? Could you try `from ..adm import CDH_Guidelines` and see which methods/properties you're missing to have parity with before?

python/pdstools/adm/Reports.py
```diff
@@ -0,0 +1,417 @@
...
+        finally:
+            if not keep_temp_files and temp_dir.exists() and temp_dir.is_dir():
+                shutil.rmtree(temp_dir, ignore_errors=True)
+
+    def _get_output_filename(self, name, report_type, model_id, output_type):
+        """Generate the output filename based on the report parameters."""
+        name = name.replace(" ", "_") if name else None
+        if report_type == "ModelReport":
+            if not model_id:
+                raise ValueError("model_id is required for a model report.")
```
StijnKas, line 214: Model ID is a required argument to this function? I don't understand how it could be missing. Sidenote: let's annotate these arguments

yusufuyanik1: It was indeed unnecessary, removed this part.

yusufuyanik1 (approved):

PR #266, opened 2024-10-15
Add generate_excel and fix mypy warnings

yusufuyanik1:
- Added generate.excel function to export 3 datamart tables. [model_data, predictor_data(aggregated) and optional full predictor_data from last snapshot]
- removed kwargs from plots
- fixed some mypy warnings

### DIFF
--- a/python/pdstools/adm/Reports.py
+++ b/python/pdstools/adm/Reports.py
@@ -353,7 +353,7 @@ class Reports(LazyNamespace):
             "Quarto executable not found. Please ensure Quarto is installed and in the system PATH."
         )
 
-    def excel(
+    def excel_report(
         self, name: Union[Path, str] = Path("Tables.xlsx"), predictor_binning=False
     ):
         """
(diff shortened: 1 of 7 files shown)


python/pdstools/adm/Reports.py
```diff
@@ -345,3 +352,56 @@ def _find_quarto_executable(self):
         raise FileNotFoundError(
             "Quarto executable not found. Please ensure Quarto is installed and in the system PATH."
         )
+
+    def excel(self, file: Path = Path("Tables.xlsx"), predictorBinning=False):
```
StijnKas, line 356: Should we just call this excel or maybe `generate.excel_export()`?

yusufuyanik1: Renamed it. Should we merge it to keep things in sync?

StijnKas:
this looks fine to me, let's get it merged @yusufuyanik1 ?

StijnKas (approved):

PR #267, opened 2024-10-21
avoid failing the report if one chart fails

yusufuyanik1:
Added try-except on 2 charts that failed with too many configurations.

### DIFF
--- a/python/pdstools/reports/HealthCheck.qmd
+++ b/python/pdstools/reports/HealthCheck.qmd
@@ -1253,26 +1253,31 @@ univariate predictor performance.
 #     else "Performance"
 # )
 
-if datamart.predictorData is not None:
-    figs = datamart.plotPredictorPerformance(
-        top_n=30,
-        facets="Configuration",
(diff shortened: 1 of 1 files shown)


operdeck:
> This is a quick workaround for the bug: https://agilestudio.pega.com/prweb/AgileStudio/app/agilestudio/bugs/BUG-893503. Just added a try except to skip the failing plots (Predictor Importance & Predictor Count). We can implement a solution to make the charts actually work but it needs a bit of thinking because I think if we allow the report to include 50 charts for each analysis it won't render. Currently it hardly renders even though we skip these 2 charts. Should we just merge this and release a patch to not block Abhijith @StijnKas ?

Makes sense to merge. I rather have the HC doc fail gracefully - but preferably with some (red) errors highlighting the issue.

BTW have you considered using Quarto's (or pandoc) column capabilities instead of plotly facets? I'm using them somewhere in one of the reports already.

StijnKas (approved):

PR #271, opened 2024-10-24
Prediction Studio library changes

patil-sharanu:


StijnKas (changes requested):
Some pretty minor comments to align with the rest of the lib (relative imports mostly) but I think this is close to ready to merge!

python/pdstools/infinity/resources/prediction_studio/v24_2/prediction.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+class Prediction(PredictionPrevious):
+    """
+    The `Prediction` class provide functionality including retrieving notifications, models,
+    adding conditional models, getting champion challengers, metrics, and plotting metrics.
+    """
+
+    @overload
+    def get_notifications(
```
StijnKas, line 27: Should we call this `list_notifications` just like we have `list_models` and `list_predictions`?

PR #272, opened 2024-10-24
fix app

yusufuyanik1:
- Add query argument to hc and model reports
- make save_data function of Datamart return abs paths
- Update health_check app code, compatible with v_4
- Update read_ds_export docstring
- make _apply_query more robust
- commented out freezing AGBTree tests.


StijnKas (approved):
Let's get to fixing the ADMTrees tests at some point, but I agree with ignoring it for now.

PR #281, opened 2024-11-08
Cosmetics to improve way reports are generated

operdeck:


operdeck:
Tests pass locally and on one of the python versions, consider the rest flakiness

PR #282, opened 2024-11-14
Small things to improve the HC. Dropped a few plots as well.

operdeck:


operdeck:
Tests work, after some re-kicks

Issue #278, opened 2024-11-08
Move Wiki contents to documentation


StijnKas:
### Is your feature request related to a problem?

We currently have documentation in two places: Wiki and documentation. 

The wiki was useful when we had two different documentation types to maintain (R and Python), but now that we've deprecated R, this is not necessary.

### Describe the solution you'd like

The upside of the documentation is that it is fully versioned and tracked in git. So, we can accept documentation PRs and include them in the docs straight away. For the wiki, we need write access to the repo that we don't want to provide to outside contributors. 

### Describe alternatives you've considered

-

StijnKas:
Closed by #283, just needs the health check article to be added. @yusufuyanik1, could you pick this up? I'll spin it up into its own issue

PR #283, opened 2024-11-18
Update documentation

StijnKas:
One of the main remaining items for V4 is/was documentation. In this PR, I've setup some infra changes to make publishing new documentation pieces easier and also added some documentation to certain classes and methods.

Additionally, with the goal of moving away from the wiki (for various reasons), this PR contains some of the getting started code that used to be in the wiki, directly in the Python docs.

I've also moved and shifted around some of the main init import logic and made certain classes not available top-level. Let's see if that gives us any issues down the line.


StijnKas:
Pretty much closes #278

PR #299, opened 2024-12-13
Light IH support focusing on Conversion monitoring

operdeck:


python/pdstools/utils/cdh_utils.py
```diff
@@ -576,9 +576,9 @@ def _capitalize(fields: Union[str, Iterable[str]]) -> List[str]:
     return fields
 
 
-def _polars_capitalize(df: F) -> F:
+def _polars_capitalize(df: F, extra:Optional[List[str]]=[]) -> F:
```
StijnKas, line 579: Ouch - lists in function signatures is asking for trouble ;)

operdeck:
@StijnKas  @yusufuyanik1  see this, first cut of implementing the IH class, plots and aggregates. Use case is monitoring conversion results, should be more general purpose however. Have a look, sprinkle in comments, and yes I'll add tests and doc later.

python/pdstools/utils/cdh_utils.py
```diff
@@ -474,7 +473,7 @@ def auc_to_gini(auc: float) -> float:
     return 2 * safe_range_auc(auc) - 1
 
 
-def _capitalize(fields: Union[str, Iterable[str]]) -> List[str]:
+def _capitalize(fields: Union[str, Iterable[str]], extra:Optional[List[str]]=[]) -> List[str]:
```
StijnKas, line 476: Same as previous comment - initializing a list here means the list remains shared across every invocation of this function

python/pdstools/ih/Plots.py
```diff
@@ -8,4 +14,130 @@
...
+            fig.add_trace(trace1, row=(r + 1), col=(c + 1))
+            index = index + 1
+
+        return fig
+
+    def tree_map(
+        self,
+        experiment_field: str,
+        by: Optional[List[str]] = None,
+        positive_labels: List[str] = None,
```
StijnKas, line 107: I think this should be type hinted as `Optional[List[str]]` too - None is not a list type

python/pdstools/ih/Plots.py
```diff
@@ -8,4 +14,130 @@
...
+        super().__init__()
         self.ih = ih
+
+    def experiment_gauges(
+        self,
+        experiment_field: str,
+        by: Optional[str] = "Channel",
+        positive_labels: Optional[List[str]] = None,
+        negative_labels: Optional[List[str]] = None,
+        reference_values: Optional[Dict] = None,
```
StijnKas, line 26: Worth type hinting the internal structure of the Dict type? Might make the API easier to use

python/pdstools/ih/Aggregates.py
```diff
@@ -8,4 +10,71 @@
...
 class Aggregates(LazyNamespace):
     def __init__(self, ih: "IH_Class"):
+        super().__init__()
         self.ih = ih
+
+    def summary_by_experiment(
+        self,
+        experiment_field: str,
+        by: Optional[List[str]] = None,
+        positive_labels: List[str] = None,
```
StijnKas, line 20: Same as before - should probably be `Optional[List[str]]` too

examples/ih/ih_helper.py
```diff
@@ -0,0 +1,87 @@
...
+import random
+import polars as pl
+from pdstools.utils import cdh_utils
+
+# Some day will move into a proper IH class
+
+# ih.plots.gauge(conversion/engagement) etc
+# constructor define objective (conversion and engagement) labels (positives/negatives)
+
+class interaction_history:
```
StijnKas, line 11: This is old I guess, right?

examples/ih/Example_IH_Analysis.ipynb
```diff
@@ -4,7 +4,19 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
```
StijnKas, line 8: should we remove this?

examples/ih/ih_helper.py
operdeck: Yes...good catch

PR #301, opened 2024-12-18
(patch): removed Plotly prerelease, added timestamp, increased DX client timeout

StijnKas:


StijnKas:
Installation was getting tricky because UV doesn't automatically allow prereleases, and it caused other libraries like pandas and pyarrow to be missed, and we still had some to_pandas in some places... let's just tie to an older version, and fix this later.

PR #308, opened 2024-12-30
polars expression serialization

yusufuyanik1:
## Changes
1. Fixed Polars expression serialization for healthcheck filters. 
   - Previously: Polars automatically converted expressions to JSON format
   - Now: Explicitly serialize expressions before writing to params.yml
   - Why: Newer Polars versions default to binary format which can't be safely stored in YAML
   - Solution: Added proper serialization/deserialization handling for expressions. Serialize just before creating YAML and deserialize in heatlhcheck.qmd

2. Code Organization Improvements
   - Simplified executable version handling as per @operdeck's TODO suggestion
   - Moved utility functions (find_quarto_executable) from Reports.py to report_utils.py
   - Resolved circular import issue where Reports class was being imported in its util file

### DIFF
--- a/python/pdstools/adm/Plots.py
+++ b/python/pdstools/adm/Plots.py
@@ -137,7 +137,7 @@ def add_bottom_left_text_to_bubble_plot(
                 else:
                     print(fig.data, i)
         return fig
-    num_models = df.select(pl.first().count()).collect().item()
+    num_models = df.select(pl.first().len()).collect().item()
     bottomleft = get_nonperforming_models(df)
     newtext = f"{num_models} models: {bottomleft} ({round(bottomleft/num_models*100, 2)}%) at (50,0)"
     fig.layout.title.text += f"<br><sup>{newtext}</sup>"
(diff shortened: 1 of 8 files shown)


yusufuyanik1:
@operdeck 
match-case statements are failing in python 3.9 because they are introduced in 3.10
will change them to if-elif-else if there is no objection

```{python}
def rag_background_styler(
    rag: Optional[Literal["Red", "Amber", "Yellow", "Green"]] = None,
):
    match rag[0].upper() if len(rag) > 0 else None:
        case "R":
            return style.fill(color="orangered")
        case "A":
            return style.fill(color="orange")
        case "Y":
            return style.fill(color="yellow")
        case "G":
            return None  # no green background to keep it light
        case _:
            raise ValueError(f"Not a supported RAG value: {rag}")
 ```

operdeck:
Stupid python but yes no objections :) Thanks

Op ma 30 dec 2024 om 16:33 schreef yusufuyanik1 ***@***.***>

PR #311, opened 2025-01-07
Hc fixes

yusufuyanik1:
- Fixed predictor_data in Datamart init. It used to not filter out models which are not in model_data @StijnKas any problems with that?
- Added active/inactive info to excel report
- Added warning to excel when the data exceed excel row limit
- Fixed predictors_overview function, used to have a query parameter but the parameter wasn't used. Now the function doesn't have query parameter, instead i filter datamart itself before calling this function
- In Health Check some maturity_criteria filters were wrong because in cdh_limits performance is in 50-100 range and we were using that to filter Performance in datamart which is in 0.5-1.0 range.


StijnKas:
I am not sure about the join behavior on models in the predictor data and not in the model data. Let's discuss on this before merging

PR #321, opened 2025-01-23
Over time

yusufuyanik1:
- Now we properly handle showing the difference of a metric over periods.
@StijnKas can you check line 302 in Plots.py, are you happy with the y-axis labels?
- Improved the unit test

### DIFF
--- a/python/pdstools/adm/Plots.py
+++ b/python/pdstools/adm/Plots.py
@@ -265,7 +265,7 @@ class Plots(LazyNamespace):
         by: Union[pl.Expr, str] = "ModelID",
         *,
         every: Union[str, timedelta] = "1d",
-        show_changes: bool = False,
+        cumulative: bool = True,
         query: Optional[QUERY] = None,
         facet: Optional[str] = None,
         return_df: bool = False,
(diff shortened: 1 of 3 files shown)


python/pdstools/adm/Plots.py
```diff
@@ -280,8 +280,8 @@ def over_time(
             The column to group by, by default "ModelID"
         every : Union[str, timedelta], optional
             By what time period to group, by default "1d"
-        cumulative : bool, optional
-            Whether to take the cumulative value or the absolute one, by default False
+        show_changes : bool, optional
```
StijnKas, line 283: Not sure how much I like this name. I think cumulative is more technically correct - technically the distinction is between cumulative or delta. We could make it a literal between these? That's even more explicit

python/pdstools/adm/Plots.py
```diff
@@ -342,48 +343,58 @@ def over_time(
...
                 .agg(pl.sum(metric))
                 .sort(grouping_columns + ["SnapshotTime"])
             )
-        else:
+
+        if show_changes:
             df = (
                 df.with_columns(
-                    Delta=pl.col(metric).cast(pl.Int64).diff().over(grouping_columns)
+                    PeriodChange=pl.col(metric).diff().over(grouping_columns)
```
StijnKas, line 356: PeriodChange seems unnecessarily ambiguous; it's just the delta between snapshot times right? Period implies that could be something different from day-to-day changes

StijnKas (commented):
Great stuff, thanks @yusufuyanik1! My only doubt is the name for the argument - imo either cumulative or delta makes sense (as they are opposites in this case), period change seems ambiguous. Other than that - good to go

PR #324, opened 2025-03-03
Remove PyArrow Dependency

yusufuyanik1:
removed to_pandas statements


StijnKas:
Nice one @yusufuyanik1 ! Changes look good, after the tests pass, I think this one deserves a proper release with release notes rather than a patch release.

PR #329, opened 2025-03-20
Fix the 'Binning of the selected Predictor' table

yusufuyanik1:
Rounding floating points was causing the total percentage to be 100.1%. Also fixed the TODO to not sum up "z-score","lift" values
Before:
Screenshot 2025-03-20 at 17 23 31

After:
Screenshot 2025-03-20 at 17 18 42


operdeck:
Nice!

Op do 20 mrt 2025 om 17:29 schreef yusufuyanik1 ***@***.***>

PR #338, opened 2025-04-10
Easypz/us 666822/agb global explanations

EHenryPega:
Added new functionality to process Adaptive Gradient Boost SHAP explanation files and then create plots of model predictor contributions.


python/pdstools/adm/GradientBoostGlobalExplanations.py
```diff
@@ -0,0 +1,791 @@
...
+            self.conn.close()
+            exit(1)
+            
+        self.conn.close()
+
+        return {'numeric': pl.concat(dfs_numeric), 'symbolic': pl.concat(dfs_symbolic)}
+
+
+
+    def _build_where(self, context_keys: dict = dict(), predictor_names: list = list()):
```
StijnKas, line 379: Putting initializers in your function definition is asking for trouble :) (https://stackoverflow.com/a/4535674)

python/pdstools/adm/GradientBoostGlobalExplanations.py
```diff
@@ -0,0 +1,791 @@
...
+            
+            df = self.conn.execute(query).df()
+            
+            self.write_to_parquet(df, f'{predictor_type}_OVERALL_{{i}}.parquet')
+            
+        except Exception as e:
+            print(f'Failed overall for predictor type={predictor_type}, err={e}')
+            return
+        
+    def parquet_overall(self, predictor_type, where_condition = 'TRUE'):
```
StijnKas, line 282: This is a more general comment: but I'm missing better typehints here; can the WHERE_CONDITION be any string? is there a subset of values? If so, I would recommend using `Literal['TRUE', 'FALSE']`, but there's no good way to tell right now.

The same goes for arguments like `data`: are they dataframes? Json files/paths? If they're dataframes, is that Pandas or Polars? Can it be eager or only lazy frames?

StijnKas (commented):
Hey @EHenryPega - thanks for this, I like where it's going!! 

I put some comments along the way, but overall, I have four main comments:
- Let's try to find a different way to handle the new imports; can we make them optional and stop the classes from being loaded in the top-level init?
- The classes have a ton of methods, which do not have type hints in their initializers, do not have docstrings and are all public. I fear it'll be tricky for people to figure out what to call when their autocomplete gives them a load of functions to call. Does it make sense to put some of these in a dedicated namespace, like I've done for the ADMDatamart class? Or perhaps some of them could be private (with a leading underscore)? As a rule of thumb, I try to document the top-level exposed functions with docstrings & type hints, but the internal functions are typically in a different namespace and don't have to be fully documented. Docstrings in numpy style are automatically documented (e.g. https://pegasystems.github.io/pega-datascientist-tools/autoapi/pdstools/adm/ADMDatamart/index.html) so that's a nice benefit :)
- It's probably a todo, but if you can add some unit tests, we can test if they are properly picked up by our CI
- Do you want your example notebook to be picked up as an article like https://pegasystems.github.io/pega-datascientist-tools/articles/ADMBinningInsights.html? That's not very difficult to do; all of these articles are driven by notebooks already, but it is nice because it gives them a fully rendered and nice to read public URL. I can help set it up (you just need to add the path to a CI file) - and you can test it locally in your fork if you enable CI in your own repo

Oh and if you could just run https://docs.astral.sh/ruff/ over your codebase, we'll have consistent formatting with the rest of the code :)

Thanks again, looking forward to bringing this in!

pyproject.toml
```diff
@@ -57,6 +60,7 @@ all = ['pdstools[app, pega_io, onnx]']
 dev = ['pre-commit']
 docs = ['sphinx','furo','sphinx-autoapi','nbsphinx','sphinx-copybutton','myst-parser', 'sphinx-argparse']
 tests = ['testbook', 'pytest', 'pytest-cov', 'pytest-httpx', 'pytest-mock', 'moto', 'openpyxl', 'pdstools[all]', 'coverage']
+explanations = ['duckdb', 'pandas', 'pyarrow']
```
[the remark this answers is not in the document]
EHenryPega: Removed pandas no problem, unfortunately duckdb seems to depend on pyarrow in order to write to a polars dataframe. https://duckdb.org/docs/stable/guides/python/polars.html

python/pdstools/adm/GradientBoostGlobalExplanations.py
```diff
@@ -0,0 +1,791 @@
...
+    SYM_QUERY_PER_TREATMENT = "sym_data"
+    NUM_QUERY_OVERALL = "num_overview"
+    SYM_QUERY_OVERALL = "sym_overview"
+    
+    NUM_TBL_NAME = "num_data"
+    SYM_TBL_NAME = "sym_data"
+
+    def __init__(
+         self
+        ,model_name: str = ''
```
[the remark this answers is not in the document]
EHenryPega: Fixed, agreed it looks funny!

python/pdstools/adm/GradientBoostGlobalExplanations.py
```diff
@@ -0,0 +1,791 @@
...
+        
+    def create_in_mem_table(self, predictor_type=None):
+        tbl_name = self.get_table_name(predictor_type)
+        
+        sql = self.read_create_table_sql_file()
+        query = self.get_create_table_sql_formatted(sql, tbl_name, predictor_type)
+        
+        self.conn.execute(query)
+        
+    def delete_in_mem_table(self, predictor_type):
```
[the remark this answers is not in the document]
EHenryPega: I would say no, the use-case here is always reading a local parquet file into an in-memory table in order to aggregate data. Once we've aggregated we can throw away the in-mem table. There's no intention to persist any data using duck-db.

python/pdstools/adm/GradientBoostGlobalExplanations.py
EHenryPega: Fixed typing for all the method parameters and their return types.

As for the where condition, the intention here is to support some kind of custom filtering. This is written into the SQL. By giving a default of 'TRUE' we end up with SELECT * FROM <table_name> WHERE TRUE, so here the where will be ignored thanks to the optimizer. 

But in theory the where_condition facilitate any filtering based on column values, for instance for a specific set of context keys. I'm not sure if we need this functionality yet, if not I will remove it completely.

pyproject.toml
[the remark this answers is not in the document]
EHenryPega: Completely agree about not including plotly and jupyter dependencies by default. I've tried to reorganise things so that explanations is entirely optional.

I think will always need plotly for the DJS job related to explanations, as our output is going to be a report which will mainly be plots.

Interested to know if you have any more pointers as of the most recent commit?

There are still a few things left on my todo list that we will make time for in the coming sprints:

- Lift some of the more utility based methods/classes into their own namespaces.
- Decouple the code for generating plots from the ipywidgets, so they can be plotted on their own (as well as in an interactive widget).
- Make the example notebook into an html as you mentioned in your previous comment. Given that we will need to setup a report html file for DJS in the near future, I think I will wait until we do that and as they can essentially be the same report.

We have a bit of time to get the above done as well as any other improvements you can think of, as we wont want to merge this until the global explanation files are actually supported in 25.x

python/pdstools/adm/GradientBoostGlobalExplanations.py
StijnKas: The convention we normally follow is to have a `query` argument -> in the docstring you can specify that it should be valid SQL WHERE syntax; and if no `query` is supplied (i.e., default value is None), no query is applied. Would that make sense here too?

pyproject.toml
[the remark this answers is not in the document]
StijnKas: All sounds great! So just to reiterate; you'd like to hold off on merging this until '25 releases? Fine by me - just needs some rebasing by that time but I don't foresee any major changes in our codebase soon, so it shouldn't be bad.

The Infinity class is one where I've done quite a bit of magic for smooth imports, just as the namespaces for the ADMDatamart class. The ADMDatamart namespaces are probably a better and more easily applicable example specifically for these optional imports; I gave every namespace its own dedicated set of dependencies.

In your case, I could imagine you'd have a `explanations.Widgets` namespace, which has all of the Jupyter-related dependencies; but those dependencies would only be checked once you actually invoke the namespace. Is that the kind of pointer you're after? Otherwise let's just discuss in the office ;)

python/pdstools/adm/GradientBoostGlobalExplanations.py
EHenryPega: Right now I'm not sure we really need to support this in the first place so it may well be removed by the time we merge. But if we do need it the approach you suggest sounds good.

pyproject.toml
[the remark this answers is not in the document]
EHenryPega: Now that I think about it we may want to merge things a bit earlier than the '25 release so we can work on the DJS job that produces the explanations report. Ideally we would be able to work against a branch instead but I'm not sure if that will be possible.

Will swing by in the office to chat about namespace magic ;)

operdeck:
I would merge spoon. Easier to keep stuff in sync and probably helpful
during dev.

Op di 15 apr 2025 om 16:03 schreef EHenry ***@***.***>

python/pdstools/explanations/ExplanationsReportGenerator.py
```diff
@@ -0,0 +1,134 @@
...
+        self._run_quarto()
+
+        if self.zip_output:
+            self.generate_zipped_report(self.report_filename, self.report_output_dir)
+
+    def _validate_report_dir(self):
+        if not os.path.exists(self.report_dir):
+            os.makedirs(self.report_dir, exist_ok=True)
+
+    def _run_quarto(self):
```
operdeck, line 71: There is a very similar run quarto in the ADM generator/reports already. In fact this whole "report generator" should be refactored so there is a common util to do this. Arguably separate from the current ADM class and into a separate utils. @StijnKas

.gitignore
```diff
@@ -35,3 +35,11 @@ python/*.ipynb_checkpoints/*
 r/tests/testthat/d/tmp2
 **/cache
 .venv
+**/_site/
+**/.tmp/
+**/.quarto/
+python/pdstools/reports/GlobalExplanations/by-context/
+.dev
+.python-version
```
StijnKas, line 43: why add .python-version here? That's a `uv` thing, I don't mind having it there :)

pyproject.toml
```diff
@@ -32,6 +34,7 @@ keywords = [
 ]
 requires-python = ">=3.9"
 dependencies = [
+    "jupyter>=1.1.1",
```
StijnKas, line 37: I would like to avoid adding Jupyter as a core dependency here. Why is that needed?

python/pdstools/explanations/ExplanationsDataLoader.py
```diff
@@ -0,0 +1,523 @@
...
+    _COL,
+    _CONTRIBUTION_TYPE,
+    _PREDICTOR_TYPE,
+    _SPECIAL,
+    ContextInfo,
+    ContextOperations,
+)
+
+
+class ExplanationsDataLoader:
```
StijnKas, line 17: Could you refactor this to be similar to the `ADMDatamart` class? In short:

1. The main class (in this case, Explanations?) is your entry point, and it has different class methods for different data sources (but could be one if you're expecting the same format). It 'holds' the data as a scanned dataframe, but does not read it
2. Plots are under a different namespace: `plots`. That way, you can do `adm.plot.bubble_chart()`
3. Aggregations are not in the main class, but put in a similar `aggregations` namespace such that we can refer to it globally
4. Anything else, such as report generation, can also get its own namespace. That way it's always clear which namespace is responsible for what. 

I don't think 'loading' the data is complex enough to have a dedicated `loader` class because Polars does that heavy lifting for us. So, I would say let's just have an `Explanations` class and have sub-namespaces under it.

python/pdstools/explanations/ExplanationsDataPlotter.py
```diff
@@ -0,0 +1,248 @@
...
+from .ExplanationsDataLoader import ExplanationsDataLoader as DataLoader
+from .ExplanationsUtils import ContextInfo, _CONTRIBUTION_TYPE, _COL, _SPECIAL
+
+logger = logging.getLogger(__name__)
+
+X_AXIS_TITLE_DEFAULT = "Contribution"
+Y_AXIS_TITLE_DEFAULT = "Predictor"
+
+
+class ExplanationsDataPlotter:
```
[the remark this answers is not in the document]
StijnKas: Agreed. So, that would look something like:

```python
class Plots(LazyNamespace):
    dependencies = ["plotly"]
    dependency_group = "explanations"

    def __init__(self, explanations: "Explanations"):
        self.explanations = explanations
        super().__init__()
```

See `python/pdstools/adm/Plots.py` for a reference.

python/pdstools/explanations/ExplanationsDataPlotter.py
```diff
@@ -0,0 +1,248 @@
...
+    
+    @classmethod
```
StijnKas, line 19: Refactoring this to be similar to the datamart variant also means all of these functions are not class methods but instance methods, so the class method decorator can be stripped and the first arg can be `self`. you then don't need the `data_loader` there anymore, because `self.explanations` will have your aggregations.

python/pdstools/explanations/ExplanationsDataPlotter.py
```diff
@@ -0,0 +1,248 @@
...
+    def plot_contributions_for_overall(
+        cls,
+        data_loader: DataLoader,
+        top_n: int = 10,
+        top_k: int = 10,
+        descending: bool = True,
+        missing: bool = True,
+        remaining: bool = True,
+        contribution_calculation: str = _CONTRIBUTION_TYPE.CONTRIBUTION.value,
```
StijnKas, line 28: Is that really a string typehint or is it an enum? Given you have a default value there, I mean

python/pdstools/explanations/ExplanationsDataPlotter.py
```diff
@@ -0,0 +1,248 @@
...
+            y_col=_COL.BIN_CONTENTS.value,
+            x_title=contribution_type.alt,
+        )
+
+        return header_fig, overall_fig, predictors_figs
+
+    @staticmethod
+    def _plot_overall_contributions(
+        df: pl.DataFrame,
+        x_col: str,
```
StijnKas, line 155: It is ultimately up to you, but I don't think these (x_col, y_col, x_title, y_title) parameters are necessary. We can set sensible defaults for them, and if the user really wants to override them they can do so on the Plotly object directly. But please weigh in here if you think there's a good reason to have them

python/pdstools/explanations/ExplanationsExplorer.py
```diff
@@ -0,0 +1,278 @@
...
+
+from .ExplanationsDataLoader import ExplanationsDataLoader
+from .ExplanationsUtils import ContextInfo, ContextOperations, _CONTRIBUTION_TYPE
+from .ExplanationsDataPlotter import ExplanationsDataPlotter
+from ipywidgets import widgets
+from IPython.display import display
+from typing import List, Optional, cast
+
+
+class ExplanationsExplorer:
```
StijnKas, line 11: What does 'explorer' mean here? I would expect a docstring explaining it, but is it really aggregations? More like EDA? This should probably also be a namespace for the main Explanations class - perhaps the name 'explorer' makes sense, but I don't know

python/pdstools/explanations/ExplanationsUtils.py
```diff
@@ -0,0 +1,158 @@
...
+
+class _SPECIAL(Enum):
+    REMAINING = "remaining"
+    TOTAL_FREQUENCY = "total_frequency"
+    MISSING = "missing"
+
+ContextInfo = TypedDict("ContextInfo", {"context_key": str, "context_value": str})
+
+
+class ContextOperations:
```
StijnKas, line 82: Should probably also be a namespace on the main Explanations class?

python/pdstools/explanations/GradientBoostGlobalExplanations.py
```diff
@@ -0,0 +1,465 @@
...
+from typing import Optional
+
+from .resources import queries as queries_data
+from .ExplanationsUtils import _PREDICTOR_TYPE, _TABLE_NAME, _COL
+from .ExplanationsReportGenerator import ExplanationsReportGenerator
+
+logger = logging.getLogger(__name__)
+
+
+class GradientBoostGlobalExplanations:
```
StijnKas, line 22: Is this a namespace for the main Explanations class, or the other way around?

Code here looks good, just wondering where it fits in the overall hierarchy of classes

python/pdstools/reports/GlobalExplanations/_environment
StijnKas, line 1: Should this be included?

python/pdstools/reports/GlobalExplanations/scripts/generate_report.py
```diff
@@ -0,0 +1,210 @@
...
+# and the overview file
+TEMPLATES_FOLDER = "./assets/templates"
+INTRODUCTION_FILENAME = "getting-started.qmd"
+OVERVIEW_FILENAME = "overview.qmd"
+ALL_CONTEXT_HEADER_TEMPLATE = "all_context_header.qmd"
+ALL_CONTEXT_CONTENT_TEMPLATE = "all_context_content.qmd"
+SINGLE_CONTEXT_TEMPLATE = "context.qmd"
+
+
+class ReportGenerator:
```
StijnKas, line 28: Is this duplicated with the ExplanationsReportGenerator?

StijnKas (commented):
Some additional comments from my side. I think my main question is the overall structuring of these classes - I would prefer if we could structure them similarly to the way we do with the ADMDatamart class where we have one main class that 'holds' the lazy data. Meaning, initializing that is 'free'. Then, it has different namespaces that actually 'do stuff' -> those namespaces could have optional dependencies, and do the actual work. 

Doing it that way makes it really easy to update things down the line because there's one obvious place new additions/changes should slot in.

Happy to give more context if needed.

python/pdstools/explanations/ExplanationsUtils.py
shaniyahassanali: Could you elaborate more? Not sure I get what you mean. It is not used in the Explanations class

python/pdstools/explanations/ExplanationsExplorer.py
shaniyahassanali: open to suggestions on the name which is more intuitive. But this class is having all the operations related to the interactive widgets that display the plots. Have a look at the notebook for examples which will give you an idea of how it works

python/pdstools/explanations/ExplanationsDataPlotter.py
shaniyahassanali: these values are coming from the calling method `plot_contributions_by_context` or `plot_contributions_for_overall` where we can pass a contribution calculation type. Based on it, the values are set as well as aggregations computed. If they are not set, the default type is "Contribution" and default calculation is also normal contribution. Ofc y_axis_title is not changing, it is always `Predictor` but for consistency passing both.

pyproject.toml
shaniyahassanali: probably got added when running locally the venv. will be removed

operdeck:
Closed as superseded by a new PR per request from Shaniya

PR #344, opened 2025-04-23
Improved time selection

operdeck:
ADM & Predictions both can summarize on start-end dates. Made fields more consistent.


operdeck:
New Action count introduced as well as options to do a summary for a specific date range.

PR #362, opened 2025-05-22
Cast omnichannel to floats

StijnKas:
@operdeck, I noticed that these summaries produce a Null column for OmniChannel if there is no omnichannel information available. Even though the values should remain Null, having the column be a Null types makes the files incompatible for direct glob parquet scanning (as the schema doesn't match).

You can decide whether to merge this, but I think this is better :)


operdeck (approved):
Good catch

PR #385, opened 2025-07-15
Make normalize_all_binnings lazy

StijnKas:
@operdeck , the binning normalization was created in the init of the class, meaning the binning normalization was _always_ run, irrespective of whether it was actually used. Turning it into a cached property should make it lazy, so it will only run when it is accessed. Should make initializing the ADMDatamart class much faster and lazy.


operdeck:
Thanks. I was vaguely aware of a slow down, well spotted!

Op di 15 jul 2025 om 14:48 schreef Stijn Kas ***@***.***>

PR #388, opened 2025-07-16
Add standard NBAD configuration to summary_by_configuration

StijnKas:
@operdeck -> will this break anything on your end that may not be caught by CI? I think it's useful to have here in addition to knowing AGB models

### DIFF
--- a/python/pdstools/adm/Aggregates.py
+++ b/python/pdstools/adm/Aggregates.py
@@ -571,9 +571,11 @@ class Aggregates:
                 .group_by(grouping)
                 .agg(
                     pl.col("AllActions").unique(),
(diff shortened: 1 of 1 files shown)


operdeck:
Likely something in the x-customer scripts but we’ll fix as it comes. We
should just review how the configuration summary table looks in the HC.

Op wo 16 jul 2025 om 16:00 schreef Stijn Kas ***@***.***>

StijnKas:
Makes sense. Since all tests pass, let's merge and if we need to make additional cosmetic changes in the HC let's fix them separately?

operdeck (commented):
I'll have a quick look at the HC after merge

PR #395, opened 2025-07-22
Don't default `pdstools` CLI -> let user choose between health_check and decision_analyzer

StijnKas:
@yusufuyanik1 , @operdeck, any concerns with this? I would like to ultimately be able to run `uvx pdstools` and have you choose between health check and decision analyzer -> then you don't even need to pre-install pdstools, just uv. Please test this out to see if it works like you would like.

### DIFF
--- a/python/pdstools/cli.py
+++ b/python/pdstools/cli.py
@@ -22,13 +22,12 @@ def create_parser():
         "app",
         choices=["health_check", "decision_analyzer"],
         help='The app to run: "health_check" or "decision_analyzer"',
-        default="health_check",
-        nargs="?",  # This makes the 'app' argument optional, allowing the default to take effect
+        nargs="?",  # This makes the 'app' argument optional
     )
     run_parser.set_defaults(func=run)
 
     # Set 'run' as the default subcommand
-    parser.set_defaults(command="run", func=run, app="health_check")
+    parser.set_defaults(command="run", func=run)
 
     return parser
 
@@ -40,7 +39,6 @@ def main():
     # Manually handle the default command if none is provided
     if args.command is None:
         args.command = "run"
-        args.app = "health_check"
 
     args.func(args, unknown)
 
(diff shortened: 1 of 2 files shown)


python/pdstools/cli.py
```diff
@@ -22,13 +22,12 @@ def create_parser():
         "app",
         choices=["health_check", "decision_analyzer"],
```
operdeck, line 23: Rather than the nerdy lowercased underscored names, why not use "Health Check" and "Decision Analyzer"?

python/pdstools/cli.py
```diff
@@ -40,16 +39,40 @@ def main():
     # Manually handle the default command if none is provided
     if args.command is None:
         args.command = "run"
-        args.app = "health_check"
 
     args.func(args, unknown)
 
 
 def run(args, unknown):
```
operdeck, line 46: Why is this run method both in py.typed and cli.py? Looks pretty identical.

yusufuyanik1:
I am a bit confused, it is giving me "No module named streamlit" error even after installing it
```
uyany@L103034MEU pega-datascientist-tools % uv pip install streamlit
Audited 1 package in 31ms
uyany@L103034MEU pega-datascientist-tools % uvx pdstools            
Traceback (most recent call last):
  File "/Users/uyany/Library/Caches/uv/archive-v0/PIUKuqEOjpvgcTtYUCxIp/bin/pdstools", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/uyany/Library/Caches/uv/archive-v0/PIUKuqEOjpvgcTtYUCxIp/lib/python3.12/site-packages/pdstools/cli.py", line 45, in main
    args.func(args, unknown)
  File "/Users/uyany/Library/Caches/uv/archive-v0/PIUKuqEOjpvgcTtYUCxIp/lib/python3.12/site-packages/pdstools/cli.py", line 49, in run
    from streamlit.web import cli as stcli
ModuleNotFoundError: No module named 'streamlit'
uyany@L103034MEU pega-datascientist-tools % 
```

python/pdstools/cli.py
StijnKas: py.typed is a completely empty file, no? it just indicates to mypy that the code is typed, doesn't actually contain anything. where are you seeing anything?

python/pdstools/cli.py
StijnKas: so that you don't have to quote these arguments when giving them as command line arguments (i.e., `pdstools run decision_analyzer` vs `pdstools run "Decision Analyzer"`. We still need that shortcut for programmatic use - the 'options' thing is just for convenience. but perhaps we could use an alias when choosing from options?

operdeck: An alias sounds good

PR #399, opened 2025-07-31
Adding Gradient Boosting Global Explanations

shaniyahassanali:


python/pdstools/utils/quarto_utils.py
```diff
@@ -0,0 +1,90 @@
...
+            if destination_path == "":
+                    destination_path = "./"
+            
+            if os.path.isdir(source_path):
+                shutil.copytree(source_path, destination_path, dirs_exist_ok=True)
+            else:
+                shutil.copy(source_path, destination_path)
+                
+    @staticmethod
+    def run_quarto(
```
operdeck, line 30: This and similar code has moved from the adm folder into report_utils. Please refactor to use that common code. This quarto_utils is probably no longer needed.

python/pdstools/explanations/Explanations.py
```diff
@@ -0,0 +1,118 @@
...
+        self._set_date_range(from_date, to_date)
+
+        self.batch_limit = batch_limit
+        self.memory_limit = memory_limit
+        self.thread_count = thread_count
+        self.progress_bar = progress_bar
+
+        self.aggregates = Aggregates(explanations=self)
+        self.data_loader = DataLoader(explanations=self)
+        self.plots = Plots(explanations=self)
```
operdeck, line 78: Pls follow suit, in the rest of PDS tools the namespace is "plot" 
e.g. self.plot = Plots(datamart=self)
you can of course name the class to whatever you like (Plots is just fine)

python/pdstools/utils/quarto_utils.py
shaniyahassanali: yes, you can decline the old pr. this one is rebased with latest master. There are no unit tests in the previous pr either, added them now

python/pdstools/explanations/FilterWidget.py
```diff
@@ -0,0 +1,269 @@
...
+
+        # init widget objects
+        self._combobox_widgets: dict[str, widgets.Combobox] = {}
+        self._selector_widget = None
+
+        self._selected_context_key = None
+
+        super().__init__()
+
+    def interactive(self):
```
StijnKas, line 36: worth a little docstring for what this does, where the filters get applied, whether it re-computes when you click something, etc.

python/pdstools/explanations/FilterWidget.py
```diff
@@ -0,0 +1,269 @@
...
+            return False
+        return all(
+            value != self._ANY_CONTEXT for value in self._selected_context_key.values()
+        )
+
+    def get_selected_context(self):
+        """Get the currently selected context information."""
+        return self._get_selected_context(with_any_option=False)
+
+    def set_selected_context(self, context_info: Optional[ContextInfo]):
```
StijnKas, line 117: Like the interactive one, probably worth a docstring

python/pdstools/explanations/Plots.py
```diff
@@ -0,0 +1,302 @@
...
+                    top_k=top_k,
+                    descending=descending,
+                    missing=missing,
+                    remaining=remaining,
+                    contribution_calculation=contribution_type.value,
+                )
+            )
+
+            for plot in [context_plot, overall_plot] + predictor_plots:
+                display(plot)
```
StijnKas, line 61: Rather than using display, which is Jupyter specific, wouldn't it make more sense to use `fig.show()`, which is native plotly? That way you can still use this outside of Jupyter.

Also, it may be a consideration to return the plot objects too, so that you can save them somewhere if you want

python/pdstools/explanations/Plots.py
```diff
@@ -0,0 +1,302 @@
...
+                descending=descending,
+                missing=missing,
+                remaining=remaining,
+                contribution_calculation=contribution_type.value,
+            )
+
+            for plot in [overall_plot] + predictor_plots:
+                display(plot)
+
+    def plot_contributions_for_overall(
```
StijnKas, line 80: would it not be easier to maintain having the functional code for this one and the 'by context' function directly in the `contributions` plot?

python/pdstools/explanations/Plots.py
```diff
@@ -0,0 +1,302 @@
...
+        predictors_figs = self._plot_predictor_contributions(
+            df_predictors,
+            x_col=contribution_type.value,
+            y_col=_COL.BIN_CONTENTS.value,
+            x_title=contribution_type.alt,
+        )
+
+        return overall_fig, predictors_figs
+
+    def plot_contributions_by_context(
```
StijnKas, line 133: same as above

python/pdstools/explanations/Plots.py
```diff
@@ -0,0 +1,302 @@
...
+
+    X_AXIS_TITLE_DEFAULT = "Contribution"
+    Y_AXIS_TITLE_DEFAULT = "Predictor"
+
+    def __init__(self, explanations: "Explanations"):
+        self.explanations = explanations
+        self.aggregate = self.explanations.aggregate
+        super().__init__()
+
+    def contributions(
```
StijnKas, line 34: Could you document this method? What does it do/plot? What do the arguments mean?

python/pdstools/explanations/Preprocess.py
```diff
@@ -0,0 +1,378 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+if TYPE_CHECKING:
+    from .Explanations import Explanations
+
+
+class Preprocess(LazyNamespace):
+    dependencies = ["duckdb", "polars", "importlib_resources"]
```
StijnKas, line 25: Pretty sure importlib_resources is base python, no?

python/pdstools/explanations/Preprocess.py
```diff
@@ -0,0 +1,378 @@
+__all__ = ["Preprocess"]
+
+import pathlib
+import os
+import duckdb
+import logging
+import polars as pl
```
StijnKas, line 7: Could you run Ruff over all these files so it properly sorts the imports? Base modules would then be sorted separately from installable

python/pdstools/explanations/Preprocess.py
```diff
@@ -0,0 +1,378 @@
...
+        self.data_foldername = "aggregated_data"
+        self.data_folderpath = pathlib.Path(
+            os.path.join(self.explanations.root_dir, self.data_foldername)
+        )
+
+        self.from_date = explanations.from_date
+        self.to_date = explanations.to_date
+        self.model_name = explanations.model_name
+
+        self.batch_limit = int(os.getenv("BATCH_LIMIT", "10"))
```
StijnKas, line 45: Thanks for making these environment variables. Could you document them in the Explanations class?

python/pdstools/explanations/Reports.py
```diff
@@ -0,0 +1,104 @@
...
+        self.aggregate_folder = self.explanations.aggregate.data_folderpath
+        self.params_file = os.path.join(self.report_folderpath, "scripts", "params.yml")
+        super().__init__()
+
+    def generate(
+        self,
+        report_filename: str = "explanations_report.zip",
+        top_n: int = _DEFAULT.TOP_N.value,
+        top_k: int = _DEFAULT.TOP_K.value,
+        zip_output: bool = False,
```
StijnKas, line 47: If zip_output defaults to False, why is the default filename .zip?

(also, please document this method)

python/pdstools/utils/report_utils.py
```diff
@@ -70,8 +71,8 @@ def _write_params_files(
         )
 
 def run_quarto(
-    qmd_file: str,
-    output_filename: str,
+    qmd_file: str = None,
```
StijnKas, line 74: Doesn't Ruff yell at you when you do this? If default is None, then the type is either `Optional[str]` or `str | None`. I think we try to use `Optional[str]` wherever we can

python/tests/explanations/test_Explanations.py
```diff
@@ -0,0 +1,108 @@
...
+
+class TestExplanationsInit:
+    """Test the initialization of the Explanations class"""
+
+    def test_default_initialization(self):
+        """Test initialization with default parameters"""
+        explanations = Explanations()
+
+        # Test default values
+        assert explanations.root_dir == ".tmp"
```
StijnKas, line 18: I think we can delete the tests that just test value initialization

.gitignore
```diff
@@ -35,3 +35,6 @@ python/*.ipynb_checkpoints/*
 r/tests/testthat/d/tmp2
 **/cache
 .venv
+**/.tmp/
+.dev
+.python-version
```
StijnKas, line 40: `.python-version` can stay in the repo, doesn't have to go in the gitignore

python/pdstools/explanations/Reports.py
shaniyahassanali: the filename is for the output zipped file, it is not used otherwise as we are creating a website using quarto. Use case for djs is the output filename will be also passed from infinity as it needs to be shown on the ui for users to know which file is generated (also downloadable). The filename is generated from infinity and send on.

python/pdstools/utils/report_utils.py
shaniyahassanali: no? 😅 it hasn't raised anything. Will add it

python/pdstools/explanations/Plots.py
shaniyahassanali: the report generation plots both overall and context level plots. both are required which are called as needed in the report generation. See [here](https://github.com/pegasystems/pega-datascientist-tools/pull/399/files#diff-9eb83a258f91abe198dee0b3c0b19aa6d2bdf211ccc22e3f88bd39884bd5b7fdR4) and [here](https://github.com/pegasystems/pega-datascientist-tools/pull/399/files#diff-c4e462b3c532c9c49c128a452a17f2d12e4e43ee85b18f0191a20aacf514dcccR13)

PR #408, opened 2025-09-23
global explanations - efficient report generation

shaniyahassanali:
Changes to render quarto report generation in batches instead of single file for all actions. Getting quarto OOM exception when using a single file with large number of actions.


operdeck:
@shaniyahassanali @yusufuyanik1  tests are failing but maybe because of something else than the SHAP changes - pls have a look

yusufuyanik1:
The error is unrelated. HTTP 403 Forbidden errors when trying to download the California housing dataset from scikit-learn for the prediction studio/onnx tests.

I think we should fix it by mocking the data or adding it to our data folder depending on the size. But it should be done on a separate PR.
@operdeck

StijnKas (approved):

yusufuyanik1:
The ONNX test is fixed (switched the dataset), but there is another test failing s3, boto3 related

shaniyahassanali:
@StijnKas / @operdeck , can we merge this pr if no further comments? the unit test failing seems to be unrelated again -> test_get_s3_files

PR #426, opened 2025-10-17
update global explanations article

yusufuyanik1:
- Added the AGB Explanations notebook to articles, 5 minutes after merging this, we should see it in: https://pegasystems.github.io/pega-datascientist-tools/index.html
- Added an **Optional** `data_file` parameter that takes a path to a single explanation file (URL or local path). This lets us read the data from github url, which is used in Github actions to run the notebook and render it in Articles.
- Changed the notebook name and wrote it in snake_case

This does not distrupt the way we read the files from a local folder, so I don't expect any problem with how it runs in DJS.

fyi @shaniyahassanali, @operdeck

### DIFF
--- a/python/tests/explanations/test_ExplanationsPreprocess.py
+++ b/python/tests/explanations/test_ExplanationsPreprocess.py
@@ -14,8 +14,9 @@ from pdstools.explanations.Preprocess import Preprocess
 
 basePath = Path(__file__).parent.parent.parent.parent
 
+
 def clean_up(root_dir):
-    _root_dir = Path(f'{basePath}/{root_dir}')
+    _root_dir = Path(f"{basePath}/{root_dir}")
     if _root_dir.exists():
         for file in _root_dir.iterdir():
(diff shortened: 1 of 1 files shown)


python/pdstools/explanations/Preprocess.py
```diff
@@ -440,6 +452,25 @@ def _populate_selected_files(self):
...
         self.selected_files = files_
 
+    def _populate_selected_files_from_url(self, file_url: str):
+        from ..pega_io.File import read_ds_export
+        import tempfile
+
+        logger.debug("Downloading file from %s", file_url)
+
+        base_path, filename = file_url.rsplit("/", 1)
+        temp_dir = pathlib.Path(tempfile.mkdtemp()) / "explanations"
```
shaniyahassanali: Can we maybe use the same root and explanations dir as in explanations.root_dir and explanations.data_folder? Keeps all the files in the same root folder

Issue #140, opened 2023-10-11
Linter/formatter CI/CD


StijnKas:
### Is your feature request related to a problem?

We don't have any linting and/or formatting as part of CI/CD

### Describe the solution you'd like

We should have linting/or formatting as part of CI/CD

### Describe alternatives you've considered

-

### Additional context

Having a good linter would assert formatting, assert easy-to-fix bugs and code smells, and would allow us to cover things like https://github.com/pegasystems/pega-datascientist-tools/pull/134#discussion_r1354896591

PR #444, opened 2025-11-03
Add Linter 

jmelburg:
Added ruff linter workflow.
#140


StijnKas:
> works but finds quite some format issues, I can tackle them in this PR as well

Well - looks like it works indeed! Nice one.

PR #467, opened 2025-11-19
HC speedup

operdeck:
ADM Health Check generation started to create very large files.
Main culprit seems to be the box plots from plotly express that include all the raw data. These are replaced by graph objects that only contain the aggregates necessary.
Addressed a few minor other issues along the way


operdeck:
Test passing, HC up again however resulting size not reduced as I expected

PR #468, opened 2025-11-20
CLI lets user choose between health check and decision analyzer

### LINKED ISSUE #465: CLI improvements
@jmelburg , is this something you can try to look at? The use case here is that it would be great if we could simply run `uvx pdstools` which would ensure you have python and the app dependencies installed, and run the health check or decision analyzer app. We could default to health check, give the user an option, or think of something else. I tried a first stab in this PR but I never got to completing it.

_Originally posted by @StijnKas in https://github.com/pegasystems/pega-datascientist-tools/issues/395#issuecomment-3521417713_

jmelburg:
supersedes #395 
resolves #465 

the streamlit import error is due to it not being included in the core dependencies. You have to run this:
```bash
uvx 'pdstools[app]'
```

If this is acceptable I think it now all works as expected, though I can't properly test it with uvx before we publish a new version.
For now I was testing with:

```bash
uv tool install -e '.[app]'
uvx pdstools
```

I'm not sure `uvx pdstools` can work out of the box without making streamlit a core dependency.


StijnKas:
Pretty sure we could make both of these work at the same time, right? 

Let's document these two in the documentation and call it a day :)

PR #476, opened 2025-11-27
Fix trend chart

yusufuyanik1:
@operdeck this PR solves all the problems we encountered on the new dataset

Items I focused on:
- change sample data in the app to use the same source as the article, used to point to sharepoint, now points to github/data, proof that no one used the app in their local except us..., or ignored the error on the Home page of the app.
- Explain the trend chart in the app better and fix it.
- make win-loss and personalization pages more robust to nothing winning and empty filter situations
- In reRank function fill V-C-L values with 1 if they are null. This breaks the arbitration as Priority becomes null and sort function doesn't behave well with null values.

### DIFF
--- a/python/pdstools/decision_analyzer/decision_data.py
+++ b/python/pdstools/decision_analyzer/decision_data.py
@@ -501,7 +501,7 @@ class DecisionAnalyzer:
     def getDistributionData(
         self,
         stage: str,
-        grouping_levels: str | List[str],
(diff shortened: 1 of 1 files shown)


operdeck:
Thanks! I love solutions that solve all problems.

Op do 27 nov 2025 om 19:08 schreef yusufuyanik1 ***@***.***>

PR #508, opened 2026-01-21
Performance thresholds changed to percentages

jmelburg:
Performance (AUC) was written as an absolute number, but is interpreted as a percentage later on. This is straightened out now.

### DIFF
--- a/python/tests/test_number_format.py
+++ b/python/tests/test_number_format.py
@@ -118,7 +118,7 @@ class TestMetricFormats:
     @pytest.mark.parametrize(
         "metric_id,value,expected",
         [
-            ("ModelPerformance", 0.875, "0.88"),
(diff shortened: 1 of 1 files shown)


jmelburg:
After MetricLimits.csv thresholds were updated from whole numbers (52, 55, 90, 100) to fractions (0.52, 0.55, 0.9, 1.0), RAG coloring in the health check was incorrect:

- Performance values like 55, 70 were compared against thresholds 0.52, 0.55, 0.9
- Almost everything showed RED because 55 > 0.9 (max threshold)

__Fix__: Normalize Performance data from 50-100 → 0.5-1.0 during data loading to match thresholds. The numbers are still presented as whole percentages in the health check document. Potentially something we should straighten out once and for all in the future.

PR #608, opened 2026-03-24
feat(impactanalyzer): add statistics module — CI, significance & sample size planning

shreyab375:
## What this adds

Significance testing, confidence intervals, and sample-size planning for Impact Analyzer experiments.

When a user sees "2% lift" in the Impact Analyzer, they need to know:
1. **Is this lift statistically significant?** — `result.significant`
2. **What's the margin of error?** — `result.ci` (95% CI)
3. **How many impressions do I need before reading the experiment?** — `required_sample_size()`

## Usage

```python
from pdstools.impactanalyzer import (
    calculate_engagement_lift,
    calculate_value_lift,
    required_sample_size,
)

result = calculate_engagement_lift(
    accepts_test=4124, impr_test=41855,
    accepts_control=81, impr_control=843,
)
print(result.lift)        # 0.0254485282
print(result.ci)          # 0.2144
print(result.significant) # False

n = required_sample_size(baseline_rate=0.05)
print(n)  # required total impressions
```

### `statistics.py`
| Function | Description |
|---|---|
| `accept_rate()` | Accept rate (Accepted + Clicked) / Impressions |
| `binomial_ci()` | Normal-approximation CI for a proportion |
| `calculate_lift()` | Relative lift: (test − control) / control |
| `lift_pl()` | Polars expression for lift (used by `summarize_experiments`) |
| `error_propagation()` | Delta-method CI for the lift ratio |
| `is_significant()` | True when CI does not cross zero |
| `calculate_engagement_lift()` | Full engagement lift with CI and significance |
| `calculate_value_lift()` | Value-per-impression lift with CI |
| `required_sample_size()` | Two-proportion z-test sample size |
| `LiftResult` | Dataclass holding lift, CI, significance, and rates |

### `ImpactAnalyzer.py`
- `summarize_experiments()` now uses `lift_pl()` from `statistics.py` (centralised lift formula)

### Tests (80 total)
- Pure formula tests with known values
- 16 PDC validation tests (4 experiments × 4 channels)
- 10 Infinity validation tests (engagement lift, value lift, CI — from a Scenario Planner Actuals export)
- 18 VBD validation tests (3 experiments × 6 channels)
- Edge cases and robustness

All formulas validated against Pega Infinity for parity.


operdeck:
> Every formula was reverse-engineered from Pega's Java implementation (ExperimentMetrics.java, ConfidenceLevelUtils.java) and validated 20/20 exact match against real PDC output.

How do we know? Afaik we don't send all the raw data to PDC that we need for validation, so we really should be validating against Infinity, using a scenario planner data export. I don't mind merging the code before we have that but I think we can't claim Pega compatibility until we have done this. @shreyab375

operdeck (commented):
I’d like to request some changes before merge so this is aligned with pdstools standards and product positioning.

**1. Frame this as core Impact Analyzer capability**
Please avoid presenting this as an add-on statistics module.
Significance testing, confidence intervals, and sample-size guidance are integral to Impact Analyzer interpretation and should be positioned as core functionality.

**2. Remove customer-identifying labels and references (blocking)**
This repository is open source, so customer-identifying names in tests, comments, ids, or assertion messages **cannot be included**.
Please replace all such labels with generic, non-identifying names.

**3. Tighten validation wording**
The PDC parity work is meaningful however, we should avoid implying full validation against real Infinity behavior until we also validate against Scenario Planner Actuals exports.
Suggested wording: validated for PDC parity; Scenario Planner Actuals validation is still pending.

**4. Keep scope and API framing aligned with pdstools conventions**
The implementation quality is promising, but please keep the public framing and change scope consistent with existing project patterns so the change remains maintainable.

operdeck (commented):
### Suggested inline comment locations

| File | Line(s) | Point |
|---|---|---|
| `python/pdstools/impactanalyzer/__init__.py` | L2-12 (the `from .statistics import` block) | Point 1: trim exports |
| `python/pdstools/impactanalyzer/statistics.py` | L246 (`def lift_pl(...)`) | Point 2: missing return type |
| `python/tests/test_ia_statistics.py` | L633-640 (`TestRequiredSampleSize`) | Point 3: use exact assertions |
| `python/pdstools/impactanalyzer/statistics.py` | L88 (module docstring, Java refs) | Point 4: rephrase internal refs |

---

## Review comment

Good work addressing round 1 -- customer names removed, Infinity validation added, lift formula centralized. The statistical implementation is solid and the Pega-parity validation is thorough. A few things to tighten before merge:

### 1. Trim the public API surface

The `__init__.py` re-exports 10 new symbols. For something that should stay clean and simple, most consumers only need the composite functions + dataclass:

- **Keep exported:** `calculate_engagement_lift`, `calculate_value_lift`, `required_sample_size`, `LiftResult`
- **Remove from `__init__.py`:** `accept_rate`, `binomial_ci`, `calculate_lift`, `error_propagation`, `is_significant`, `lift_pl`, `Z_95`

These primitives are implementation details. `lift_pl` is only used internally by `ImpactAnalyzer.py` -- it can import directly from `.statistics` rather than going through the package public API. Same for `Z_95` -- a constant shouldn't be in the package's `__all__`.

This keeps `from pdstools.impactanalyzer import ...` clean and avoids committing to a public contract on internal helpers.

### 2. Missing return type annotation on `lift_pl`

```python
def lift_pl(test_col: str, control_col: str):  # <- missing -> pl.Expr
```

Per our coding guidelines: "Use type hints for public APIs and complex internal functions." Since polars is lazy-imported inside the function, you can use a string annotation or `from __future__ import annotations`.

### 3. Some test assertions are weaker than they could be

Our guidelines strongly prefer exact-value assertions over structural checks. A few tests only verify relative ordering or bounds:

- `TestRequiredSampleSize`: `test_returns_positive`, `test_higher_baseline_needs_fewer`, `test_smaller_mde_needs_more` -- these only check `n > 0` or `n_a > n_b`. With known inputs and a deterministic formula, we can assert exact expected values.
- `test_large_sample` (`ci < 0.002`), `test_very_large_numbers` (`ci < 0.02`), `test_tiny_control_group` (`ci > 0.5`) -- directional bounds, not exact.

These are fine as supplementary smoke tests, but adding the exact computed values alongside would strengthen them and catch regressions that stay within the bounds but are still wrong.

### 4. Minor: module docstring references internal Java classes

The `statistics.py` docstring mentions `ExperimentMetrics.java` and `ConfidenceLevelUtils.java`. Since this is an open-source repo, consider rephrasing to "Formulas follow Pega's server-side implementation" without naming internal source files.

operdeck (commented):
Lets add some visuals, show error bands or something in the graphs. Make our UI support the same analyses and metrics as product - but not vis in the same way, much more straight-forward: per experiment, value & engagement lift, error bands on those etc. Perhaps even show the formulae used - not sure about that.

StijnKas (changes requested):
## Review

Thanks for the depth on the formulas, the tile-based result layout, and the test coverage — those parts genuinely shine. The statistical maths is correct, and 112 exact-value tests against PDC / Infinity / VBD exports is the kind of due diligence the rest of the repo should aspire to.

But architecturally, this PR is going in the wrong direction. I think it needs a substantial rework before merge. Apologies for the length of what follows — the maths is great so a lot of the work is salvageable, but the structure needs to change.

---

### Where I think this needs to land

`pdstools` has a core design principle that isn't currently written down (which is partly our fault — we'll fix that): **all functionality lives in the library. Streamlit apps are zero-functionality presentation layers.** Anything an app can do — every calculation, every transformation, every rendering of a formula — must be reproducible in a Jupyter notebook or a script using only `pdstools.<module>` imports. Apps add accessibility (UI, no coding required); they never add capability. If a feature only exists in the app, it's misplaced.

By that yardstick, this PR puts most of its value in the wrong place. The good news: the maths and the layout are reusable. The work is reorganising, not rewriting.

---

### 🔴 Blocker 1 — `ia_engine.py` is a parallel re-implementation of `statistics.py`

The same PR that introduces `pdstools.impactanalyzer.statistics` ships an `examples/.../ia_engine.py` that re-implements `Z_95`, `_binomial_se`, `_lift`, `_lift_se`, `_significant`, validation logic, and experiment definitions. Future divergence is guaranteed: someone fixes a bug in `statistics.lift` and the engine version silently drifts.

The example app — the most prominent demonstration of the new module — currently demonstrates the wrong pattern.

**Required:** delete `ia_engine.py`. The app must call the public `pdstools.impactanalyzer.statistics` API. If the API doesn't fit (e.g. the app wants per-event validation), the API needs to grow — *in the library* — not be forked into the app.

---

### 🔴 Blocker 2 — The transparency content belongs in the existing IA app, not a parallel one

We already have `python/pdstools/app/impact_analyzer/` with a `Home.py`, navigation, pages, and full launcher integration via `pdstools run impact-analyzer`. It uses `standard_page_config`, `show_sidebar_branding`, the shared theming, and the existing session-state conventions.

This PR ignores all of it and ships a parallel app at `examples/impact-analyzer-transparency-app/` with its own `st.set_page_config()`, its own colour tokens (`PEGA_BLUE`, `PEGA_GREEN`, …), its own sidebar layout, and a separate launch story. Two apps doing similar things is a worse experience than one app with two pages.

**Required:** the transparency content lands as **a new page (or pages) inside `pdstools.app.impact_analyzer`**, reusing `standard_page_config`, the existing theming, and the existing session state conventions. The example directory shouldn't ship a Streamlit app at all — see the next point.

---

### 🔴 Blocker 3 — The "examples" directory should demonstrate the library, not host the product

What we'd actually want under `examples/impact-analyzer/` is a **Jupyter notebook** that walks through the same content the app shows: load a payload, call `pdstools.impactanalyzer.statistics.calculate_engagement_lift(...)`, render the formulas as markdown, plot the lift with CI, etc. That notebook proves the principle: anything the app does, a script can do too.

If the notebook can't be written — if some piece of the app calls into `ia_engine.py` and not into `pdstools` — that's the gap to close, not a reason to keep the engine.

**Required:** `examples/impact-analyzer/` becomes a notebook (or short script) that consumes the library; the Streamlit app moves into `pdstools.app.impact_analyzer/pages/`.

---

### 🔴 Blocker 4 — Formulas need a first-class library representation

The app currently bakes formulas into Python strings inside `ia_engine.py` / `app.py` and renders them via ad-hoc `_render_formula(latex, description, substitution)` helpers. That keeps them locked inside the app.

What we want: formulas live in `pdstools.impactanalyzer.statistics` as structured objects — something like

```python
@dataclass(frozen=True)
class Formula:
    name: str
    latex: str           # raw LaTeX with placeholders like {p_test}, {n_test}
    description: str

    def filled(self, **values) -> str:
        """Return the LaTeX with values substituted in, ready to render."""
```

attached to each function (e.g. `binomial_se.formula`) or registered in a small lookup. Then:

- The library uses them for documentation / reference.
- The app calls `binomial_se.formula.filled(p=0.04, n=1000)` and passes the string to `st.latex(...)` or `st.markdown(...)`.
- A notebook calls the same thing and passes it to `IPython.display.Latex(...)`.
- Tests can assert formulas render correctly with known inputs.

The app contains zero formula text; the library contains all of it.

**Required:** model the formulas as library objects; the app becomes a renderer.

---

### 🔴 Blocker 5 — `LiftResult.ci` is a standard error, not a confidence interval

The module docstring explicitly says: *"Pega stores standard errors (SE), not confidence intervals."* But the `LiftResult.ci` field's docstring says: *"Delta-method confidence-interval half-width for lift."* The implementation stores the SE (no z-multiplier). Side-effect: `is_significant` checks `[lift ± SE]` against zero, which is a ~68% confidence test, not 95%.

External users who see `result.ci` will assume it's a 95% half-width and may multiply by 1.96 themselves, double-scaling.

**Required:** rename `LiftResult.ci` → `se`; rename `error_propagation` → `lift_se`; add a `ci_95()` convenience method (`return Z_95 * self.se`); update `is_significant` docstring to be explicit about which confidence level it tests at and why.

---

### 🔴 Blocker 6 — `build_excel` doesn't belong in the app (or in this PR)

`app.py` defines a 100+ line `build_excel(results)` that imports `openpyxl` inline and hand-builds Summary / Detail / Formulas worksheets with custom fonts, fills, borders, column widths. There's nothing like this anywhere else in `pdstools` — no other app exports Excel, and there's no library pattern for "structured tabular export with styling".

By the core principle: this is functionality, so it has to live in the library. By the "earn your abstractions" principle: we don't have a generic need for this yet — no other class wants to export Excel. So the right move is to drop it.

**Required:** delete `build_excel` from this PR. If Excel export is a real user request, file it as a separate issue; design it across `ADMDatamart` / `DecisionAnalyzer` / `ImpactAnalyzer` together; declare `openpyxl` as an optional extra; ship it with tests.

---

### 🔴 Blocker 7 — "Live bridge" mode adds a network client to an analysis library

`app.py:_poll_bridge` polls an HTTP "live bridge" endpoint via `requests`. Same principle: if it's functionality, it lives in the library, not the app. And we don't currently have any Pega live-stream client in `pdstools` — adding one for a single example is premature.

**Required:** drop the live-bridge mode, or move the polling client into a properly-scoped optional extra in `pdstools.pega_io` with documentation of the bridge protocol.

---

### 🟡 Should-fix (statistics module quality)

- **`LiftResult` mutability:** `@dataclass` → `@dataclass(frozen=True)`. Result objects shouldn't mutate.
- **`binomial_ci` docstring:** add a one-line note about normal-approximation under-coverage at extreme `p` and reference Wilson / Clopper-Pearson alternatives.
- **`required_sample_size` default `control_ratio=0.02`:** correct for IA but surprising for general power-analysis. Document in the docstring.
- **Numpy-style docstrings:** the new `statistics.py` uses prose docstrings. Per the convention now codified in [AGENTS.md → Docstring style: numpy](https://github.com/pegasystems/pega-datascientist-tools/blob/master/AGENTS.md#docstring-style-numpy), public APIs need `Parameters` / `Returns` / `Raises` sections with dashed underlines. Ruff D406/D407/D410/D411 now enforce this in CI.

---

### 🟢 What's good (please keep)

- **Statistical formulas are correct** — delta method, binomial SE, two-proportion z-test all check out.
- **Test quality** — 112 exact-value tests against real PDC / Infinity / VBD exports. Stellar.
- **`ImpactAnalyzer.py` integration** — replacing inline `_lift_pl` with the centralised `lift_pl` is clean and minimal-surface. ✅
- **Lazy `scipy` import** in `required_sample_size` — correct pattern.
- **The tile-based result layout** — visually it's the right direction, please reuse the design when porting into the existing app.

---

### Suggested next steps

I'd split this into a sequence of smaller PRs:

1. **`statistics.py` cleanup** — `LiftResult.ci → se` rename, `error_propagation → lift_se`, `frozen=True`, numpy-style docstrings, the `should-fix` items above. Land this as a tight, focused PR.
2. **Formula objects** — add the structured `Formula` representation to `statistics.py`, attach to each function, with tests that verify `.filled(...)` renders correctly. No app changes yet.
3. **Notebook example** — `examples/impact-analyzer/transparency.ipynb` walking through the same content the app shows, using only `pdstools` imports. This validates that the library is complete enough to back the app.
4. **App integration** — port the tile layout, charts, and formula rendering into `pdstools.app.impact_analyzer/pages/` consuming the library. Drop `ia_engine.py`, `build_excel`, `_poll_bridge`, the parallel colour tokens, and the parallel `set_page_config`.

The maths and the layout are reusable across all four PRs, so the work is reorganising what you already have, not redoing it.

We'll also be adding a "Streamlit apps are zero-functionality presentation layers" principle to AGENTS.md so this is documented for future contributors — the issue is on us for not having that written down before. Happy to pair on the restructuring if it's useful.

shreyab375:
## CI note: `test_upload_replaces_autoloaded_sample` failure

The `test_ia_upload_replaces_autoload.py::test_upload_replaces_autoloaded_sample` test may fail in CI. This is a **pre-existing issue from master** (introduced by PR #767), not caused by changes in this PR.

**Root cause:** The test calls `at.get("file_uploader")[0].upload(...)`, but Streamlit's `AppTest` testing framework returns an `UnknownElement` for `file_uploader` widgets — the `.upload()` method does not exist on that element type, resulting in `AttributeError: upload`.

This was verified by checking out master's `_home_page.py` against the same test — it also fails (on the duplicate About page error before even reaching the upload step, which this PR fixes).

**Changes in this push:**
- **`_navigation.py`**: Removed duplicate `3_About.py` from `_SUB_PAGES` (was also in `_ABOUT_PAGE`, causing `StreamlitAPIException: Multiple Pages specified with URL pathname About`)
- **`2_Drill_Down.py`**: Changed `st.title()` to markdown heading (`"# Drill Down"`) to match test expectations and the pattern used by `1_Overall_Summary.py`
- **`_home_page.py`**: Added autoload of sample data on first visit to match master's `ensure_ia_data_loaded()` behavior expected by tests

PR #631, opened 2026-04-13
fix: weighted average report corrected

shaniyahassanali:
# fix: weighted average report corrected

**Branch:** `team/easypz/TASK-1890691-weighted-average`  
**Commit:** `ad6f815e`

## Summary

Fixes two related bugs in the explanations/weighted-average reporting logic in the `Aggregate` class.

### Bug fix: weighted average formula

Corrects the weighted average calculation from `.mean()` to `.sum()` in the `_apply` helper. The previous formula was computing `mean(contribution × frequency) / total_frequency`, which double-averaged and produced incorrect results. The correct formula is `sum(contribution × frequency) / total_frequency`.

### Bug fix: `frequency_over` scope for weighted contributions

The frequency denominator (`total_frequency`) is now computed per predictor (grouped by `PARTITION + PREDICTOR_NAME + PREDICTOR_TYPE`) rather than per partition only. This ensures each predictor's weighted average divides by its own bin frequencies, not the entire partition's total.

### New parameter: `include_numeric_single_bin`

Moves single-bin numeric predictor filtering from data load time (always-on) to query time (opt-in via `include_numeric_single_bin=True`, default `False`). This is threaded through `get_contributions`, `get_top_bins`, `_get_weighted_average_contributions`, and all `Plots` methods.

## Files changed

- `python/pdstools/explanations/Aggregate.py` — formula fix, frequency scope fix, lazy filtering
- `python/pdstools/explanations/ExplanationsUtils.py` — adds `include_numeric_single_bin` default
- `python/pdstools/explanations/Plots.py` — threads `single_bin_numeric` parameter through all plot methods

RCA -> 
[rca-weighted-avg-presentation.html](https://github.com/user-attachments/files/26671601/rca-weighted-avg-presentation.html)

### DIFF
--- a/python/pdstools/explanations/Aggregate.py
+++ b/python/pdstools/explanations/Aggregate.py
@@ -55,6 +55,7 @@ class Aggregate(LazyNamespace):
         descending: bool = defaults.descending,
         missing: bool = defaults.missing,
         remaining: bool = defaults.remaining,
+        include_numeric_single_bin: bool = defaults.include_numeric_single_bin,
         sort_by: str = defaults.sort_by.value,
     ):
         """Get the top-n predictor contributions for a given context or overall.
@@ -71,6 +72,9 @@ class Aggregate(LazyNamespace):
                 Whether to include contributions for missing predictor values.
             remaining (bool):
                 Whether to include contributions for remaining predictors outside the top-n.
+            include_numeric_single_bin (bool):
+                Whether to include numeric predictors that have only a single
+                non-missing bin. Default is False (exclude them).
             sort_by (str):
                 Method to sort/select top contributions. Options include
                 `contribution`, `contribution_abs`, `contribution_weighted`.
@@ -93,6 +97,7 @@ class Aggregate(LazyNamespace):
             descending=descending,
             missing=missing,
             remaining=remaining,
+            include_numeric_single_bin=include_numeric_single_bin,
             sort_by=validated_sort_by.value,
         )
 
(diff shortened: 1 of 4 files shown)


python/pdstools/explanations/Aggregate.py
```diff
@@ -55,6 +55,7 @@ def get_predictor_contributions(
         descending: bool = defaults.descending,
         missing: bool = defaults.missing,
         remaining: bool = defaults.remaining,
+        include_numeric_single_bin: bool = defaults.include_numeric_single_bin,
```
operdeck, line 58: @shaniyahassanali  What is that 'include_numberic_single_bin' that I see all over the place? Why do we need this flag?

python/pdstools/explanations/Plots.py
```diff
@@ -115,6 +118,7 @@ def plot_contributions_for_overall(
         descending: bool = defaults.descending,
         missing: bool = defaults.missing,
         remaining: bool = defaults.remaining,
+        include_numeric_single_bin: bool = defaults.include_numeric_single_bin,
         sort_by: str = defaults.sort_by.value,
```
operdeck, line 122: Do all these plot methods need all those 3 flags?

operdeck (commented):
One thing to address: `include_numeric_single_bin` is threaded through 7 signatures across three files to control two lines of filtering. That kind of parameter proliferation adds API surface without user benefit — could this be handled closer to where the filtering actually happens (e.g. in the two internal methods that need it) rather than exposed through every public and private method?

The `_Defaults` dataclass pattern also leans a bit enterprise-y for pdstools. Plain keyword-argument defaults (`top_n: int = 20`) are more idiomatic and self-documenting.

We've codified both points in AGENTS.md — see the new ["Keep parameter surfaces small"](https://github.com/pegasystems/pega-datascientist-tools/blob/master/AGENTS.md#keep-parameter-surfaces-small) and ["Use Python's built-in defaults"](https://github.com/pegasystems/pega-datascientist-tools/blob/master/AGENTS.md#use-pythons-built-in-defaults) sections under Design Principles.
@StijnKas

PR #635, opened 2026-04-15
refactor: explanations public API to use

shaniyahassanali:
# Refactor explanations public API to use `**filter_kwargs`

**Branch:** `team/easypz/feat/explanations-slim-kwargs`

## Summary

Replaces explicit `sort_by`, `descending`, `missing`, `remaining` (and `display_by`)
parameters on public methods in the explanations module with `**filter_kwargs`.
Validation and default-filling are centralized in two new resolver functions
(`_resolve_agg_filter_kwargs`, `_resolve_plot_filter_kwargs`) in `ExplanationsUtils.py`.
This keeps public signatures clean while preserving full type safety on the private
methods that do the actual work.

### `ExplanationsUtils.py` — new resolver functions
- Added `_resolve_agg_filter_kwargs(**kwargs)`: validates `sort_by`, `descending`,
  `missing`, `remaining`; fills defaults; rejects unknown keys with `TypeError`.
  Returns `sort_by` as a string (for Aggregate/Polars column operations).
- Added `_resolve_plot_filter_kwargs(**kwargs)`: extends the agg resolver with
  `display_by`; returns both `sort_by` and `display_by` as `_CONTRIBUTION_TYPE`
  enum members so callers can use `.value` / `.alt` without re-validating.
- Both resolvers are exported via `__all__`.

### `Aggregate.py` — slim public signatures
- `get_predictor_contributions()` and `get_predictor_value_contributions()` now
  accept `**filter_kwargs` instead of four explicit filter parameters.
- Docstrings updated to document valid kwargs with types and defaults.
- Removed `_CONTRIBUTION_TYPE` import (no longer needed here).
- **Bug fix:** `_get_predictor_value_contributions` line 302 — changed
  `if predictors is not None or len(predictors) > 0` to
  `if len(predictors) > 0`. After `predictors = predictors or []`, the
  `is not None` check was always `True`, making the `or` a no-op.

### `Plots.py` — slim public signatures, no double validation
- `contributions()`, `plot_contributions_for_overall()`, and
  `plot_contributions_by_context()` now accept `**filter_kwargs`.
- Removed all `_CONTRIBUTION_TYPE.validate_and_get_type()` calls — the
  resolver returns both `sort_by` and `display_by` as enums.
- Pops both enums from resolved dict, converts `sort_by` back to `.value`
  before passing to aggregate methods.
- Removed unused `_CONTRIBUTION_TYPE` import.

### `Reports.py` — slim public signature, no re-validation
- `generate_report()` now accepts `**filter_kwargs` instead of explicit
  `sort_by` / `display_by` parameters.
- Both `sort_by` and `display_by` enums are popped directly from the
  resolver — no redundant `validate_and_get_type()` call needed.

### Tests — `test_ExplanationsAggregate.py`
- New `TestFilterKwargsValidation` class: unknown kwarg rejection for both
  aggregate public methods.
- New `TestFilterKwargsDefaults` class: default equivalence and override
  behaviour for both methods.

### Tests — `test_ExplanationsPlots.py`
- Added unknown kwarg rejection test for `contributions()` and
  `plot_contributions_for_overall()`.
- Added default equivalence test for `plot_contributions_for_overall()`.
- Added override behaviour test for `plot_contributions_for_overall()`.


python/pdstools/explanations/Reports.py
```diff
@@ -61,13 +60,16 @@ def generate(
...
         top_k : int
             Number of top features to include in explanations.
-        sort_by : _CONTRIBUTION_TYPE
-            Contribution type enum used for sorting and ranking data.
-        display_by : _CONTRIBUTION_TYPE
-            Contribution type enum used for display axes and report text.
         zip_output : bool
             Whether to zip the output report.
             The filename will be used as the zip file name.
+        **filter_kwargs:
```
operdeck, line 66: If this is considered a public interface, I think this **should** have explicit arguments and not the kwargs style. If you agree, that means sort_by, display_by could be reverted to explicit arguments.

shaniyahassanali: The plots and aggregate are also public methods that were changed to use kwargs, it is following the same standard here. The arguments sent to this function are also the same parameters used for plot and aggregate. Except it currently doesn't pass `remaining`, `missing`, and `descending` as those are defaulted. The reason to pass `sort_by` and `display_by` is to use the `values` for the plot during report generation and use the `text/alt` for the homepage of the report.

operdeck:
@copilot resolve the merge conflicts in this pull request

operdeck:
@copilot resolve the merge conflicts in this pull request

operdeck:
@shaniyahassanali  we apparently ended up with some merge conflicts - can you have a look?

shaniyahassanali:
@operdeck resolved, please review and approve

Issue #648, opened 2026-04-20
DecisionAnalyzer.filtered_sample reads st.session_state from library layer


StijnKas:
## Summary

`DecisionAnalyzer.filtered_sample` (in `python/pdstools/decision_analyzer/DecisionAnalyzer.py`, lines 773–797) is a **library-layer** property that imports `streamlit` and reaches into `st.session_state.get("page_channel_expr")` to apply page-level filters.

This couples the core analyser to UI runtime state, hides a non-obvious data dependency in a property access, and silently no-ops outside Streamlit. The recently merged design principles in `AGENTS.md` (under "Right-size your validation and security" / "Keep parameter surfaces small") and the long-standing rule under "Streamlit apps" guide us toward keeping UI state out of the library layer.

The pattern is currently small (one filter, `page_channel_expr`) but `docs/plans/decision-analyzer-TODO.md` proposes extending it with Issue and Group filters, which would multiply the coupling. Fixing this before that work lands keeps the cost low.

## Current code

```python
# python/pdstools/decision_analyzer/DecisionAnalyzer.py
@property
def filtered_sample(self):
    """Sample data with page-level filters applied.

    Reads filter expressions from st.session_state.page_channel_expr if available.
    Falls back to unfiltered sample if no page filters are set or not in Streamlit context.

    This property is not cached because it depends on mutable session_state.
    ...
    """
    try:
        import streamlit as st
        channel_expr = st.session_state.get("page_channel_expr", None)
        if channel_expr is not None:
            return apply_filter(self.sample, [channel_expr])
    except ImportError:
        pass  # Not in Streamlit context
    return self.sample
```

## Why this is a problem

1. **Hidden coupling.** A core analysis class secretly depends on UI runtime state. The same `DecisionAnalyzer` instance returns different data depending on the caller's context (Streamlit vs notebook vs script).
2. **Silent fallthrough.** If Streamlit isn't installed, or `page_channel_expr` isn't set, callers silently get unfiltered data. There's no signal to distinguish "no filter" from "filter unavailable".
3. **Cannot be cached.** The docstring acknowledges this: a property whose return value depends on mutable global state is essentially uncacheable. That's the smell admitting itself.
4. **Violates `AGENTS.md` design principles.** The "Streamlit apps" section says shared utilities and app-specific helpers belong in `app/`. Filter wiring sourced from `st.session_state` is app code.

## Scope of the coupling today

- **Reads** `da.filtered_sample` in 8 pages:
[...]

PR #649, opened 2026-04-20
Decouple DecisionAnalyzer from Streamlit session_state (#648)

StijnKas:
DecisionAnalyzer.filtered_sample previously imported streamlit and reached into st.session_state.page_channel_expr, coupling the core analyser to UI runtime state. Same instance returned different data in Streamlit vs notebook/script contexts and silently no-op'd when streamlit was unavailable.

Move the session_state lookup into the app layer:

- Add DecisionAnalyzer.filtered(filters) — pure library API that takes a Polars expression or list of expressions explicitly.
- Add da_streamlit_utils.collect_page_filters() — single home for gathering page-level filter expressions from session_state. New contextual filters (Issue, Group, …) only need a key appended to the tuple here; pages and library both stay unchanged.
- Update all 9 analysis pages to use da.filtered(collect_page_filters()) instead of da.filtered_sample.
- Keep filtered_sample as a deprecation-warning shim returning the unfiltered sample, so any external code or notebook caller fails loudly rather than silently dropping filters.
- Replace streamlit-mocking tests with monkeypatched session_state dicts and add coverage for the new API surface.
- Update docs/decision-analyzer-architecture.md and docs/plans/decision-analyzer-TODO.md to describe the new pattern and the (now trivial) extension path for Issue/Group filters.

Refs #648

### DIFF
--- a/docs/decision-analyzer-architecture.md
+++ b/docs/decision-analyzer-architecture.md
@@ -95,10 +95,14 @@ page's sidebar. Session state keys:
 - `page_channel_filter` — UI display value ("Any" or "Channel/Direction")
 - `page_channel_expr` — Polars filter expression or None
 
-Applied via `DecisionAnalyzer.filtered_sample` property, which reads session state
-and filters the sample. To add new contextual filters (e.g., Issue, Group), extend
-`contextual_filters()`, add corresponding session state keys, and update
-`filtered_sample` to collect all filter expressions.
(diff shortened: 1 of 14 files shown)


StijnKas:
Resolves #648

operdeck (approved):
I did a quick manual check of the decision analyzer app. It seems to still work and I'm glad to see the dependency of the core data class on streamlit gone.

PR #653, opened 2026-04-20
Refactor report APIs to use TypedDict + Unpack for shared options

StijnKas:
Both Reports.model_reports and Reports.health_check had ~14 keyword parameters each, with 8 of them (title, subtitle, disclaimer, output_dir, output_type, qmd_file, full_embed, keep_temp_files) duplicated as identical presentation/output knobs.

Introduce ReportOptions (TypedDict, total=False) and pass it via **options: Unpack[ReportOptions] (PEP 692). The call sites stay unchanged — dm.generate.health_check(title=..., output_type='pdf') still works and now gets full IDE autocomplete + static type checking on those kwargs. Method signatures are now focused on what's actually method- specific (model_ids, query, prediction, file paths, callback, only_active_predictors).

Defaults live in per-method _MODEL_REPORT_DEFAULTS / _HEALTH_CHECK_DEFAULTS dicts. _resolve_report_options() merges and validates: unknown keys raise TypeError so typos surface at runtime (TypedDict is static-only).

Note: health_check previously allowed name/title/subtitle/disclaimer/ output_dir as positional arguments. All in-repo callers (tests, app, docs) use keyword form, so this is preserved in practice while making the API consistent with model_reports.

### DIFF
--- a/python/pdstools/adm/Reports.py
+++ b/python/pdstools/adm/Reports.py
@@ -1,6 +1,5 @@
-__all__ = ["Reports"]
+__all__ = ["ReportOptions", "Reports"]
 import logging
-import os
 import shutil
 from os import PathLike
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


StijnKas:
@operdeck fyi - slight public API change here, but I think it's for the better; makes the interface a lot cleaner

PR #733, opened 2026-04-21
feat(decision-analyzer)!: route DA app through st.navigation()

StijnKas:
Replaces Streamlit's auto-discovered pages/ navigation with an explicit st.navigation() registry. Pages that need a loaded DecisionAnalyzer (Overview, Action Distribution, Action Funnel, Global Sensitivity, Win/Loss, Optionality, Offer Quality, Thresholding, Arbitration Distribution, Single Decision) are hidden from the sidebar until data is uploaded — addressing the 'visible but immediately error-out' experience the previous auto-discovery layout produced.

When no data is loaded, the Home page surfaces a 'Upload data above to unlock these analysis pages' hint listing each gated page so users still know what's coming.

Hidden pages remain reachable by URL; existing per-page ensure_data() guards apply unchanged. Page files in pages/ are untouched, so direct AppTest.from_file() calls and existing bookmarks keep working.

standard_page_config() is now idempotent — Streamlit raises StreamlitAPIException on a second set_page_config() call in the same script run, and with st.navigation() both the entry script and the routed page invoke it. Catching the exception lets the same page modules work standalone (AppTest, direct streamlit run pages/X.py) and inside the navigation router.

Closes the custom-sidebar-navigation P1 plan item.


StijnKas:
FYI @operdeck - this is a big one. we should test it before the V5 release

PR #750, opened 2026-04-22
fix: numeric single bin and frequency calc

shaniyahassanali:
# PR Summary: Fix numeric single-bin handling and frequency calculation

**Branch:** `fix/numeric-single-bin-and-frequency-calc`
**Commits:** `f153682f`, `304e835b`
**Files changed:** 4 (83 additions, 25 deletions)

---

## Problem

Two bugs in the explanations module:

1. **Single-bin numeric predictors produced `null` intervals.** The `LAG()`/`LEAD()` SQL window functions return `NULL` for the first/last row, so predictors with only one bin had no valid min/max interval.

2. **Frequency percentages were calculated incorrectly.** Frequencies were scoped per-predictor instead of per-partition, causing percentages to always sum to 100% within each predictor rather than reflecting the predictor's share of the overall model.

### SQL: single-bin fix (`numeric.sql`, `numeric_overall.sql`)
- Wrapped `LAG()` and `LEAD()` window functions with `COALESCE(...)`, falling back to the row's own `minimum`/`maximum` when there is no adjacent bin. This ensures single-bin predictors get a valid interval instead of `NULL`.

### Python: frequency calculation fix (`Aggregate.py`)
- Changed `frequency_over` in `_calculate_aggregates` from `[partition, predictor_name, predictor_type]` to just `[partition]`, so frequency percentages represent each predictor's share of the whole partition — not 100% within itself.
- Increased `frequency_pct` rounding precision from 1 to 4 decimal places.
- Added new method `add_context_frequency_pct_to_df()` that computes a context's frequency as a share of the overall (non-contextual) model by joining against `get_df_overall()`.

### Plots: hover & frequency display (`Plots.py`)
- `_build_hover_customdata()` now accepts `include_frequency` flag; the whole-model view omits the frequency row (it's always 100%).
- `_plot_overall_contributions()` uses the new `add_context_frequency_pct_to_df()` for context views, and skips frequency for the whole-model view.
- `_plot_bin_level_contributions()` now groups frequency by `[partition, predictor_name, predictor_type]` to correctly scope per-bin percentages.

## Testing notes

- Verify numeric predictors with a single bin render valid intervals (no `null`/`NaN` in tooltips).
- Verify frequency percentages in the overall contribution chart sum correctly across predictors within a partition.
- Verify hover tooltips: whole-model view should **not** show frequency; context views should show frequency as a share of the overall model.

### DIFF
--- a/python/pdstools/explanations/Aggregate.py
+++ b/python/pdstools/explanations/Aggregate.py
@@ -243,11 +243,7 @@ class Aggregate(LazyNamespace):
         # divides by that predictor's own bin frequencies, not the entire partition.
         df = self._calculate_aggregates(
             df,
-            frequency_over=[
-                _COL.PARTITON.value,
-                _COL.PREDICTOR_NAME.value,
-                _COL.PREDICTOR_TYPE.value,
-            ],
+            frequency_over=[_COL.PARTITON.value],
             aggregate_over=[
                 _COL.PARTITON.value,
                 _COL.PREDICTOR_NAME.value,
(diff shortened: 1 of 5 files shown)


StijnKas (changes requested):
## Review

Thanks for tackling these — the SQL single-bin fix and the new `add_context_frequency_pct_to_df()` method are both correct and well-scoped. There's one **blocker** below; the rest are nits.

---

### 🔴 Blocker — `frequency_over=[partition]` change in `_get_predictor_contributions()` breaks weighted averages

**File:** `python/pdstools/explanations/Aggregate.py` (around line 243-269)

The change from `frequency_over=[partition, predictor_name, predictor_type]` to `frequency_over=[partition]` looks like it was intended to fix display percentages, but it conflates two distinct concepts:

- `total_frequency` — used by the weighted-average formula `(contribution * frequency).sum() / total_frequency.first()`. **Must be per-predictor.**
- `frequency_pct` — used for display. The new `add_context_frequency_pct_to_df()` method (correctly) handles this separately.

Worked example with two predictors:

```
Age bins:   freq=100, c=0.2  +  freq=100, c=0.8   → total_per_predictor=200, weighted=0.5
Score bins: freq=10,  c=0.1  +  freq=90,  c=0.9   → total_per_predictor=100, weighted=0.82
```

With the change, both predictors divide by the partition total (300) instead of their own per-predictor totals (200 / 100), producing wrong weighted averages. The existing test `test_weighted_average_scoped_per_predictor` (lines 592-668) was added specifically to lock in the per-predictor scoping — and the inline comment on lines 265-266 explicitly says *"total_frequency is computed per predictor so the weighted average divides by that predictor's own bin frequencies, not the entire partition."* — that comment now contradicts the code.

**Suggested fix:** revert the one-line change to `frequency_over=[partition, predictor_name, predictor_type]`. The display-percentage problem is already solved by the new `add_context_frequency_pct_to_df()` method + Plots changes — no further change needed there.

---

### 🟡 Should-fix

**Add a test for `add_context_frequency_pct_to_df()`.** Frequency calculation was buggy before; lock in the new behaviour with an exact-value test that verifies `context_freq / overall_freq * 100` across multiple predictors. (See AGENTS.md → "Strongly prefer exact-value assertions".)

**Add a test for the SQL single-bin COALESCE.** Use a fixture with a single-bin numeric predictor (e.g. `include_numeric_single_bin=True`) and assert the interval renders as `[min:max]` rather than `null`.

**Justify the precision change.** `frequency_pct.round(1)` → `.round(4)` (line 599) is a behavioural change for any downstream code parsing/formatting these percentages. If the goal is to handle very small frequencies (e.g. 0.02%), a one-line note in the commit message or a code comment would help future readers.

---

### 🟢 Optional

- **SQL `[5:5]` edge case:** when all bins in a single-bin predictor have `minimum == maximum`, the COALESCE produces `[5:5]` — technically correct but might read better as `= 5`. Up to you.
- **Type hints:** `add_frequency_pct_to_df()` is missing a `-> pl.DataFrame` return annotation (`add_context_frequency_pct_to_df()` has one — nice).
- **Docstring style:** the new method uses prose-only docstrings; per the `numpy-style` convention now codified in [AGENTS.md](../blob/master/AGENTS.md#docstring-style-numpy) (#tbd-PR-link), please format `Parameters` / `Returns` sections with the dashed underlines. Going forward, ruff D406/D407/D410/D411 enforce this.

---

### Summary

Once the `frequency_over` revert is in (~one line) plus a couple of tests, this is good to merge. The other items are quality-of-life.

PR #790, opened 2026-05-11
chore: remove tracked anonymised.parquet test artifact

operdeck:
## Summary

The 4.47 KB `anonymised.parquet` file at the repo root is leftover output from `python/tests/adm/test_Anonymization.py`. The fixture used relative paths (`output_file=\"anonymised.parquet\"`, `temporary_path=\"tmp\"`), so running the test from the repo root wrote the artifact there. It was introduced once via #705 (an unrelated ADMTrees refactor) and was never intended to be tracked.

## Changes

- Remove `anonymised.parquet` from the repo.
- Patch the test fixture in `python/tests/adm/test_Anonymization.py` to use pytest's `tmp_path` for both `output_file` and `temporary_path` so the test can no longer pollute the working tree.
- Gitignore `/anonymised.parquet` (root-anchored) as belt-and-braces.

## Not changed

The library default (`output_file=\"anonymised.parquet\"`) in `pdstools/pega_io/Anonymization.py` is left as-is — fine for interactive users; the bug was only in the test.

## Verification

- `uv run pytest python/tests/adm/test_Anonymization.py -q` → 7 passed.
- `git status` clean afterwards; no new `anonymised.parquet` in the repo root.
- pre-commit hooks pass.


StijnKas:
oops. thx @operdeck

PR #804, opened 2026-05-25
TASK-1909467: Adapting prediction API across releases

chandana-telukutla:
✅ What Changed

Updated version handling logic across the API client
Refactored code to standardize version usage in API calls
Removed outdated or hardcoded version references
Improved alignment with current Prediction Studio API endpoints
Minor cleanup for better readability and maintainability

✅ Why

To support latest API versioning and endpoint structure
To ensure consistent behavior across modules
To avoid issues caused by outdated version references
To improve compatibility with Prediction Studio integration
To make future version upgrades easier


StijnKas:
Hey @chandana-telukutla - thanks for this, I've been meaning to come around to adjusting this! 

Could you let me know when this one is ready for review? I'd like to merge it fairly quickly because I want to adjust some of the internals of the API structure, but I'd prefer to do that after this one is merged. Thanks!

chandana-telukutla:
Hii @StijnKas, the PR is ready for review. Thanks!

StijnKas (changes requested):
Thanks for the PR. Supporting 25.1/26.1 is overdue and the bug fixes you've folded into v24_2 (missing leading slashes on endpoints, URL encoding for special characters, the safer OAuth error extraction) are genuinely good — those land regardless of how we resolve the rest. The new test coverage around version detection and the mutating APIs is also a real step up.

The main theme of the review: **the v25 and v26 trees are the same code twice, and the version strings don't match how Pega names releases.** Both are easy to fix and both will save us pain later.

---

## v25 and v26 are the same code twice — collapse them

The PR ships full parallel implementations of `champion_challenger`, `prediction`, `prediction_studio`, `model`, `repository`, and `datamart_export` under both `v25/` and `v26/`. Diffing the trees, the only differences are the class-name suffix (`_Modelv25Mixin` vs `_Modelv26Mixin`) and a handful of docstrings — and several of those docstrings under `v25/` already say "v26" (see `v25/champion_challenger/_mixin.py:20`, `v25/repository.py:9`, `v25/datamart_export.py:7`), which is the clearest possible signal that this is one implementation copy-pasted.

The PR's own comment in `_base_client.py:229` confirms this: *"the v25 and v26 APIs are compatible, and v26 is the default for any 25+ system"*.

What I'd suggest:

- Keep one implementation directory (`v26/`).
- Make `v25/__init__.py` a thin re-export: `from ..v26 import *  # noqa: F401, F403`.
- Update the version map in `prediction_studio/__init__.py` so both `"25.1"` and `"26.1"` resolve to the v26 classes.

That removes ~3000 lines of duplication, eliminates the drift risk (any v24_2-style bug fix that lands in v26 today will silently miss v25), and matches how `local_model_utils` and the v24 tree are already structured — one source of truth, with explicit overrides only where the API actually differs.

If a real v25-vs-v26 divergence shows up later, we can split them at that point — but ship the merged version now, not the speculative split.

## Version strings should be `"25.1"` / `"26.1"`, not `"25"` / `"26"`

The product versions are `25.1` and `26.1` — there is no 25.2 or 26.2, and the marketing/docs/release-notes are all `<major>.<minor>`. Pdstools surfacing `client.version == "26"` to users will read as wrong to anyone who's looked at a Pega release page. It also boxes us in: the day a `26.2` ships, we can't add it to the version map without breaking the "`26` means latest 26-line" assumption that's now baked into the dispatch.

What I'd suggest:

- Use `"24.1"`, `"24.2"`, `"25.1"`, `"26.1"` as the canonical keys in `prediction_studio/__init__.py`'s version map.
- Have `_infer_version` return the same canonical strings.
- If you want a forgiving lookup for users who pass `"25"`, normalise at the boundary (`version = version if "." in version else f"{version}.1"`), don't store the ambiguous form internally.

That also makes the `_infer_version` docstring honest: right now it returns `"26"` for any v25+ system, which is misleading in logs. With canonical `"26.1"` and the v25↔v26 collapse from the previous point, this becomes a non-issue — both 25.1 and 26.1 systems resolve to the same v26 resource classes, and the reported version actually matches what's running.

## Smaller things

- **Dead `None` path.** `prediction_studio/__init__.py:35-48` — `get(version)` returns `None` for falsy input, but the new `__getattr__` on the client raises `AttributeError` before ever calling `get()`. Either drop the check and tighten the return type to `type[PredictionStudioBase]`, or let `None` flow through and have `get()` own the error. Right now it's unreachable.
- **EOF newline** missing on `v26/prediction/_mixin.py:241`. Pre-commit / ruff-format will flag it on the next sweep.
- **Docstring copy-paste** — the "v26" docstrings on v25 mixins mentioned above. Moot once the trees are collapsed.

---

## Summary

To get this merge-ready:

1. Collapse `v25/` into a re-export of `v26/`; keep one implementation.
2. Switch the version map and `_infer_version` to canonical `"25.1"` / `"26.1"` strings.
3. Drop the unreachable `None` branch in `get()`.
4. Fix the EOF newline and the stray "v26" docstrings (the latter disappears with point 1).

The v24_2 fixes and the new test coverage are great and worth keeping as-is. Happy to talk through any of this if it's helpful.

chandana-telukutla:
**v25/v26 duplication**

- I’ve collapsed the implementation to a single v26/ source of truth.
- v25/ is now a thin re-export of v26/ to avoid duplication
- Updated the version map so both 25.1 and 26.1 resolve to the same v26 classes
- This removes the duplication and should prevent drift going forward

**Version strings**

- Switched everything to canonical Pega formats: "24.1", "24.2", "25.1", "26.1"
- _infer_version now returns these canonical values
- Added normalization so inputs like "25" are converted to "25.1" at the boundary

**Smaller fixes**

- Dead None path, Removed the unreachable None branch in get() and tightened the return type accordingly
- Added missing EOF newline
- Cleaned up docstrings (resolved naturally after collapsing trees)

Let me know if you’d like me to split anything further or if there are edge cases you want covered — happy to iterate.

StijnKas (approved):
Much better now, thank you @chandana-telukutla! I've pushed some small fixes and resolved some conflicts; but looks ready to merge now. I'll push it through, if there's anything you'd like to edit later just create another PR. Thanks again!

PR #833, opened 2026-06-15
feat(adm/trees): AGB Explained notebook

operdeck:
- Add Node.sample_count field to _nodes.py; _iter_nodes now yields sampleCount from the tree dict for training-timeline methods.

- Add 9 new plot methods to ADMTreesModel: plot_gain_per_tree, plot_cumulative_gain_share, plot_feature_importance_by_gain, plot_early_vs_late_gain, plot_gain_by_namespace, plot_feature_role_map (all work without sampleCount), and plot_training_stream_timeline, plot_inter_tree_gaps, plot_gain_decay_dual_lens (require sampleCount, raise ValueError without it). All follow the show/return_df/@overload pattern. Fix a bug where plot_gain_decay_dual_lens called update_xaxes(secondary_y=False) which is not a valid plotly axis property.

- Add data/agb/ModelExportWithSampleCount.json: anonymized 100-tree export (AUC 0.810, 15M responses) with sampleCount on every node. Replaces the remote-URL approach in datasets.sample_trees().

- Update datasets.sample_trees() to load from the bundled local file (local-first with path walk, matching the ImpactAnalyzer pattern).

- Add 22 new exact-value unit tests for all 9 methods covering return_df shapes and pinned values, figure return type, and ValueError on models without sampleCount.

- Extend check-added-large-files pre-commit hook to exclude data/ (consistent with existing large sample files already in that dir).


StijnKas (changes requested):
While we're refactoring this anyway - with the release of V5.0.0 we're open to breaking changes. Could we take the opportunity to turn this class into a properly namespaced one? instead of tree.plot_whatever, tree.plot.whatever.

PR #843, opened 2026-06-16
docs: reduce sphinx warning noise

StijnKas:


StijnKas:
this was too broad. trying smaller changes.

PR #844, opened 2026-06-16
fix(docs): remove legacy gh-pages content

StijnKas:


StijnKas:
cleaned up manually instead

PR #849, opened 2026-06-19
Fix: ADM HealthCheck (sub)plot and annotation formatting

jmelburg:
Solves three issues:
1. Vertically stretched plots due to increasing the number of annotations
2. Most overlapping labels (caused by the new metric limit annotations) are now separated
3. Threshold lines now appear on top of the gird lines instead of behind


jmelburg:
@StijnKas Tried with 2 large environments and all looks fine. Do you want (me) to check / review anything?

PR #860, opened 2026-06-24
Fix/metric limit names

jmelburg:
two variables in the metric limits files had names that deviate from our conventions. They started with "average" instead of ending. 

`AverageGroupsPerIssue` 

`AverageTreatmentsPerChannelPerAction`


jmelburg:
Waiting for databricks metric notebook to be unblocked by legits and racoons.

StijnKas (approved):

PR #878, opened 2026-07-10
Filter Predictor Data Before Caching

yusufuyanik1:
## Summary

Reduce ADM Health Check memory usage by pruning cached predictor data to the ModelIDs that remain after `ADMDatamart` filtering.

## Why this change

We investigated an OOM issue in `AdmHealthCheck` with a dataset.

What we observed:
- filtering to a single configuration reduced the model subset a lot, but memory stayed higher than expected

The reason is that lazy filtering on `model_data` is not enough by itself:
- `ADMDatamart` applies `query` to `model_data`
- `predictor_data` stays unfiltered
- `Reports.health_check()` was caching the full predictor snapshot before Quarto reloads it
- downstream report sections still processed predictor rows for ModelIDs that were no longer in the filtered model set

So the issue is not with lazy filtering itself; it is that the persisted report inputs did not match the filtered datamart.

## What this PR changes

Before generating the health check report, derive the filtered `ModelID`s from `datamart.model_data` and pass them to `save_data()`.

This makes the cached predictor dataset contain only rows relevant to the filtered report.

## Result

For the filtered sample case (`Configuration = <a configuration>`):

- before this change:
  - peak process-tree RSS: **5912 MB**
  - peak main Python RSS: **4567 MB**
  - Runtime: 13.4 seconds

- with this change:
  - peak process-tree RSS: **2424 MB**
  - peak main Python RSS: **1331 MB**
  - Runtime: 8.7 seconds 

Saves both memory and runtime

I don't remember we weren't doing this in the first place, but if we had a reason for it, maybe we can parameterize this behaviour and run it like this in DJS.


operdeck (changes requested):
Hey @yusufuyanik1 thanks. I don't remember either why we did not do this earlier, and lets not introduce a flag if not needed. But I have another comment - the gist of this seems to be that you consolidate the data before self.save_data. So this is covered:

dm = ADMDatamart.from_ds_export(..., query=pl.col("Configuration") == "A")
dm.generate.health_check()

But this path seems not covered (the query argument directly in the health_check): dm.generate.health_check(query=...)

The selected model IDs should probably be done earlier, before collecting selected_model_ids.

In GPT's own words:

This optimization works when the datamart was already filtered before calling `health_check()`, because `self.datamart.model_data` already contains only the selected `ModelID`s.

However, `health_check()` also has its own `query` argument. In that path, the new code collects `selected_model_ids` from `self.datamart.model_data` before applying the report query. That means `save_data()` still receives all model IDs and writes the full predictor dataset to the temp cache. The Quarto report applies the query later when it reloads the cached files, but the memory-heavy cache write has already happened.

Could we apply the `health_check()` query before collecting `selected_model_ids`, and add a small test that verifies `health_check(query=...)` passes only the matching model IDs to `save_data()`?

yusufuyanik1:
> Hey @yusufuyanik1 thanks. I don't remember either why we did not do this earlier, and lets not introduce a flag if not needed. But I have another comment - the gist of this seems to be that you consolidate the data before self.save_data. So this is covered:
> 
> dm = ADMDatamart.from_ds_export(..., query=pl.col("Configuration") == "A") dm.generate.health_check()
> 
> But this path seems not covered (the query argument directly in the health_check): dm.generate.health_check(query=...)
> 
> The selected model IDs should probably be done earlier, before collecting selected_model_ids.
> 
> In GPT's own words:
> 
> This optimization works when the datamart was already filtered before calling `health_check()`, because `self.datamart.model_data` already contains only the selected `ModelID`s.
> 
> However, `health_check()` also has its own `query` argument. In that path, the new code collects `selected_model_ids` from `self.datamart.model_data` before applying the report query. That means `save_data()` still receives all model IDs and writes the full predictor dataset to the temp cache. The Quarto report applies the query later when it reloads the cached files, but the memory-heavy cache write has already happened.
> 
> Could we apply the `health_check()` query before collecting `selected_model_ids`, and add a small test that verifies `health_check(query=...)` passes only the matching model IDs to `save_data()`?

Good catch, DJS uses this pattern:
```dm = ADMDatamart.from_ds_export(..., query=pl.col("Configuration") == "A") dm.generate.health_check()```
but I missed the pattern where query is passed to HC directly.

https://github.com/pegasystems/pega-datascientist-tools/pull/878/changes/fce1e07287b05481f497cdec92bbb32bedf5b0d1 should fix it

PR #879, opened 2026-07-10
Leaner explanations

shaniyahassanali:
## Summary

This PR streamlines the Explanations module to a leaner architecture by removing legacy preprocessing/aggregation file flow and consolidating validation and report path handling.

### What changed

1. Removed preprocessing/aggregating-file based flow and associated legacy assets/queries, simplifying data handling in `pdstools.explanations`.
2. Refactored Explanations internals (`Aggregate`, `Explanations`, `ExplanationsUtils`, `Reports`, `Plots`) to reuse validation paths and reduce duplication.
3. Standardized path handling around `pathlib.Path` in reports and related components.
4. Updated report templates and global explanations report generation scripts to match the refactor.
5. Updated parameters and docstrings for clearer API behavior.
6. Refreshed and aligned Explanations test coverage across module, plots, reports, and generation flows.
7. Synced `pyproject.toml` to reflect the refactor/test updates.

### Impact

- Cleaner and more maintainable Explanations code paths.
- More consistent path/type handling.
- Better-aligned documentation and tests for the new flow.


operdeck (commented):
Big shift from runtime aggregation to pre-aggregated data. Loads of files changed :). Just a few cosmetics in docs and enums you may want to fix before we merge

## PR879 Review Findings

`origin/master` exposed `Explanations.from_local_directory(...)` as the migration path for filesystem loading. This branch removes it and only exposes `from_aggregates`, while the migration guide and changelog still tell users to call `from_local_directory`.

Impact: users following the documented migration path get:

```python
AttributeError: type object 'Explanations' has no attribute 'from_local_directory'

**Comment by StijnKas** [comment]
Thanks for the PR. The direction makes sense: removing the legacy preprocessing stack, consolidating around pre-aggregated parquet inputs, and simplifying the report flow is all good cleanup. The API surface is definitely leaner than before.

Before we merge, though, there are a few correctness and migration issues to fix. The biggest one is that the new path handling is not wired through consistently, and a couple of the migration/docs updates now describe behavior the code no longer supports.

---

## Preserve aggregate paths end-to-end

The new `Explanations` API explicitly allows callers to point at a custom aggregate folder via `from_aggregates(root_dir=..., data_folder=...)`, including nested relative paths and absolute paths. That part is good.

The problem is that the report flow drops path information later on:

- `Explanations._init_state()` now splits custom paths into `root_dir` + `data_folder`
- `Reports._set_params()` writes only `self.aggregate_folder.name`
- `ReportGenerator._read_params()` reconstructs the folder as `../{data_folder}`

That only works when the aggregate folder is a direct child of the root. For something like:

```python
Explanations.from_aggregates(
    root_dir="/tmp/work",
    data_folder="nested/aggregated_data",
)
```

the report generator will look in `/tmp/work/aggregated_data`, not `/tmp/work/nested/aggregated_data`.

Please keep the full relative aggregate path intact through the report params, rather than collapsing it to `.name`.

## Fix the published migration story

The migration guide currently tells users to do this for Quarto/template use cases:

```python
explanations = Explanations()
explanations.aggregate.data_folderpath = DATA_FOLDER
```

That no longer matches the implementation. Validation and report generation are driven from `root_dir` / `data_folder` on the parent `Explanations` object, not from a later mutation of `aggregate.data_folderpath`. So this documented migration path leaves:

- `validate_data_folder()` pointing at the default `.tmp/aggregated_data`
- `Reports.aggregate_folder` still bound to the original path
- `ContextOperations.unique_contexts_file` still bound to the original path

In other words: the guide describes a pattern that looks plausible but is now broken. The docs should be updated to show the new supported entry point only, i.e. configuring the aggregate path via `from_aggregates(...)` / constructor arguments rather than by mutating sub-objects afterward.

## Align filenames in code and docs

The refactor renames the overall aggregate file to `OVERVIEW.parquet`, and the implementation now hardcodes that name when loading the overall dataset.

However, the migration guide and changelog still tell users to provide:

```text
BY_CONTEXT.parquet
OVERALL.parquet
```

That is now wrong and will send users straight into a runtime failure. The docs need to match the new contract everywhere this file layout is described.

## Remove the stale `plots.contributions(...)` migration example

The migration guide still uses `plots.contributions(...)` as its example for the kwargs cleanup. But `Plots.contributions()` was removed entirely in this PR.

So the example currently teaches users to migrate onto an API that no longer exists. Please swap that snippet to one of the remaining public methods (`plot_contributions_for_overall`, `plot_contributions_by_context`, `Aggregate.get_predictor_contributions`, etc.).

## Clean up the smaller API/doc mismatches

There are a few smaller inconsistencies worth fixing while you're here:

1. `Explanations` says the constructor "takes no filesystem paths", but the constructor now explicitly accepts `root_dir` and `data_folder`.
2. `ContextOperations.get_df(..., with_partition_col=True)` still documents the raw column as `partition`, but the new public name is `context_partition`.
3. `ContextInfo` and the `get_list()` docs no longer reflect the actual returned dictionaries once `with_partition_col=True` is used.
4. `ReportGenerator._get_unique_contexts()` has a dead fallback branch: after the early return and JSON-load branch, `if not self.contexts:` can only raise.

None of these are major on their own, but together they make the new API harder to trust because the code and the docs are drifting in small but visible ways.

---

## Summary

To get this merge-ready:

1. Preserve the full aggregate folder path through report param generation and report loading; don't collapse it to `Path.name`.
2. Fix the migration guide so it only documents supported path-configuration patterns.
3. Update all docs/changelog references from `OVERALL.parquet` to `OVERVIEW.parquet`.
4. Replace the stale `plots.contributions(...)` migration example with a method that still exists.
5. Clean up the remaining constructor/context doc mismatches and the dead `_get_unique_contexts()` fallback branch.

shaniyahassanali:
@StijnKas 

On `Preserve aggregate paths end-to-end`: I am unclear on the issue, with your given example it would still reconstruct the `root_dir` from the `data_folder`.

StijnKas:
> @StijnKas
> 
> On `Preserve aggregate paths end-to-end`: I am unclear on the issue, with your given example it would still reconstruct the `root_dir` from the `data_folder`.

The issue is specifically the **explicit `root_dir` + nested relative `data_folder`** case.

With this call:

```python
Explanations.from_aggregates(
    root_dir="/tmp/work",
    data_folder="nested/aggregated_data",
)
```

the constructor **does not** split `data_folder` and recover `root_dir` from it, because `_init_state()` only does that for:

1. absolute paths, or
2. multi-part relative paths when `root_dir` is still the default `.tmp`

So after `_init_state()` you have:

```python
self.root_dir == "/tmp/work"
self.data_folder == "nested/aggregated_data"
```

That part is fine.

The problem comes later:

1. `Aggregate.data_folderpath` becomes `/tmp/work/nested/aggregated_data`
2. `Reports._set_params()` writes only `self.aggregate_folder.name` → `"aggregated_data"`
3. `ReportGenerator._read_params()` reconstructs the path as `os.path.abspath(os.path.join(report_folder, "..", self.data_folder))`

So the report generator ends up using:

```python
/tmp/work/aggregated_data
```

instead of:

```python
/tmp/work/nested/aggregated_data
```

So the bug is **not** in `Explanations._init_state()` reconstructing the wrong root. The bug is that the report handoff later **collapses the full aggregate path to its final path segment** via `.name`, which loses `"nested/"`.

shaniyahassanali:
@operdeck @StijnKas can you review and merge this pr if no further comments

PR #885, opened 2026-07-15
Fix Health Check report chart layout issues

operdeck:
## Summary
- Improve Health Check trend chart layout by syncing success-rate facets and preserving model-technique facet labels when metric limit annotations are hidden.
- Add spacing fixes for faceted bubble charts, trend charts, and treemap overview titles.
- Restore malformed guidance callouts to proper Quarto callout blocks.
- Add focused regression coverage for the affected plot helper behavior.


operdeck:
I might add a few more cosmetic improvements after scanning more HC results

PR #886, opened 2026-07-16
Improve Health Check import workflow

operdeck:
Overhauls the ADM Health Check app's data import experience and fixes several related rough edges.

## Import UX

Flexible upload: zipped CSVs, faster CSV preview parsing, simpler timestamp handling, quoted path inputs.
Auto-load sample data on first visit; the banner now names the loaded source (sample / imported / configured path) instead of a stale "Import Successful!".
Show generated report and processed-parquet paths on Reports actions.

## Predictor categories

Fill/fallback null categories on load; pin standard category colors (incl. Primary distinct from IH).
Dedicated fallback palette so custom categories are visually distinct, with no hard limit on category count.

## Polish

Reworded several UI strings for clarity; muted inactive predictor-count bars; kept predictor-count colors by type.
Skip full-embed batch reports when esbuild is unavailable; test-isolation and demo-URL fixes.


python/pdstools/pega_io/File.py
```diff
@@ -226,7 +230,16 @@ def _json_pxresults_fallback(file: str | Path | BytesIO) -> pl.LazyFrame:
     return pl.from_dicts(json.loads(raw)["pxResults"]).lazy()
 
 
-def _read_from_bytesio(file: BytesIO, extension: str) -> pl.LazyFrame:
+def _read_from_bytesio(
```
StijnKas, line 247: Not a huge fan of these reading options being separate; can we not make this a generic 'reading_options' or 'polars_reading_options' and passing to whatever importer we use? Makes the signature leaner and since we only ever read in one filetype at a time there should never be any conflicts. We can/should maybe validate or leave it to Polars to produce invalid argument errors?

operdeck: Thanks. Will have a look.

PR #904, opened 2026-07-27
Refactor - Pythonic explanations & stricter linter

StijnKas:
# Refactor `pdstools.explanations` to match the rest of the library

**This is a breaking API change**, covered in `docs/migration-v4-to-v5.md`.

## Bugs fixed along the way

These are behavioural, not cosmetic:

- **Contexts past row 100 were silently dropped.** `pl.from_dicts()` defaults to
  `infer_schema_length=100`. Any context key that first appeared later in the
  file was discarded without warning — the aggregation just produced fewer
  columns than the data contained. Now `infer_schema_length=None`. (Note the
  asymmetry that made this invisible: `str.json_decode()` *raises* on the same
  drift, `from_dicts()` returns quietly.)
- **Stale cache.** Cached frames were not invalidated when the underlying
  aggregates changed, so a second call could return data from the first.
- **Missing bins were mishandled**, producing wrong contribution totals.
- **Plot output was non-deterministic** — figure/trace order varied between runs.
- **Four tests asserted nothing.** They called the method under test and then
  asserted on a literal, so they passed regardless of the result.

### Found by a post-refactor review pass

Three models reviewed the branch independently; these came out of that:

- **The default `output_dir` did not work.** `generate()` stored the *relative*
  `.tmp/reports/data` in `params.yml`. The Quarto pre-render script runs with
  the report folder as its cwd and resolves a relative `data_folder` against
  that folder's *parent*, yielding `<cwd>/.tmp/.tmp/reports/data`. Every test
  passed an absolute `tmp_path`, which is exactly why it was invisible; the
  regression test uses a relative path.
- **`_join()` broke its own contract.** It checked whether the *base* was a URL
  before checking whether the *filename* was absolute, so a URL base plus an
  absolute local filename produced `https://host/agg//abs/f.parquet`.
- **Re-running into the same output directory left orphans.** Stale
  `BATCH_<n>.parquet` files and by-context `.qmd` pages from a previous run
  survived, contradicting the documented "rewritten on every call" behaviour.
- **An empty contextual frame raised `StructFieldNotFoundError`** from
  `json_decode` returning a Null dtype. An empty dataset is legitimate, so it
  now yields no contexts.
- **`_add_total_frequency_to_df` could return a duplicated column.** Not
  reachable from current call sites, but the suffixed `total_frequency_right`
  it would produce is a silent-wrong-value trap.

### Pure `__init__`, I/O in classmethods

The constructor took `root_dir` / `data_folder` / `data_file` and ran the
aggregation pipeline as a side effect. It now takes two `pl.LazyFrame`s and
plain config; all reading lives in `from_aggregates`, whose signature mirrors
`ADMDatamart.from_ds_export`:

```python
exp = Explanations.from_aggregates(base_path=".tmp/aggregated_data")
```

A consequence worth calling out: because the class no longer holds an input
path, **the aggregates can live anywhere polars can read from**, including
behind a URL. That was structurally impossible before — one attribute served as
both the input folder and the report output folder, so reading remotely and
writing locally could not be expressed.

```python
exp = Explanations.from_aggregates(base_path="https://example.com/aggregates")
```

Report artifacts (`unique_contexts.json`, `batches/`) used to be written back
into the *source* folder. They now land in the report's own working directory
via a new `Explanations.save_data()`, following the same cache-and-pass-to-QMD
path `ADMDatamart` already uses.

### `ExplanationsUtils` dissolved

334 lines of unrelated static methods, held together by nothing but the word
"Utils". Split into `_constants.py`, `ContextOperations.py`, and methods on the
classes that own the behaviour.

### Naming aligned with the rest of the library

| Before | After |
| --- | --- |
| `Explanations.aggregate` / `Aggregate` | `Explanations.aggregates` / `Aggregates` |
| `plot.plot_contributions_for_overall(...)` | `plot.contributions_overall(...)` |
| `aggregate.get_predictor_contributions(...)` | `aggregates.predictor_contributions(...)` |
| `aggregate.get_df_overall()` | `Explanations.overall` |

The frames are now plain attributes (`exp.overall`, `exp.contextual`), mirroring
`dm.model_data` / `dm.predictor_data`.

### Other cleanups

- **`Schema.py`** — schema applied on read, as `ADMDatamart` does.
- **`**filter_kwargs` catch-alls replaced with explicit parameters** on the
  `Plots`, `Aggregates` and `Reports` public methods.
- **`validate_data_folder()` removed** — it duplicated what the reader already
  raises. A missing file now raises `FileNotFoundError` from `from_aggregates`
  itself rather than from the first operation that touches the data.
- **`create_unique_contexts_file()` + `create_batch_parquet_files()` merged**
  into one `write_batches(target_dir)`, which now streams via
  `sink_parquet(pl.PartitionBy(...))` — the contextual frame is never collected
  into memory.
- **`_set_date_range()` → pure `_resolve_date_range()`**. The old version
  mutated `self` from inside a helper; making it pure also removed an
  unreachable guard in `Reports`.
- **`root_dir` dropped.** It only ever fed the report output folder, so it is
  now a call-time argument: `exp.report.generate(output_dir=...)`.

## New: `sample_explanations()`

Follows the same pattern as `cdh_sample()` / `sample_value_finder()`, pulling
pre-aggregated data from the repo over HTTPS:

```python
from pdstools import sample_explanations
exp = sample_explanations()
```

This also fixes the explanations article notebook, which loaded aggregates via
a relative repo path. `data/` is not shipped in PyPI releases, so the notebook
was unrunnable for installed users — the problem #899 identified. It now goes
through `sample_explanations()`, and I verified the notebook executes
end-to-end. `sample_explanations` takes `from_date` / `to_date` so the article
keeps its fixed reporting window instead of defaulting to "the last seven days".

## Changes outside `explanations`

`pega_io` gained `is_url()`, and `scan_parquet_path` now accepts URLs and raises
`FileNotFoundError` for missing local paths. **This affects all `pega_io`
callers, not just explanations** — worth a look during review.

## Notes for the reviewer

- One `# pyrefly: ignore` is added with an inline explanation. It is a genuine
  pyrefly bug in constrained-TypeVar substitution (pyright is clean on the same
  code); I confirmed it with a 4-line repro. The alternative was contorting the
  signature to dodge a type-checker limitation, which seemed worse.
- The `@overload` pair on `_add_total_frequency_to_df` was replaced by the
  repo's shared `F = TypeVar("F", pl.DataFrame, pl.LazyFrame)` — that is what a
  constrained TypeVar is for.
- The lint-enforcement work that came out of this review was split into #903 and
  is already merged into this branch.

## Verification

- 2777 tests pass (145 in `explanations`)
- mypy, pyright, pyrefly all clean
- `pre-commit run --all-files` clean
- Full CI matrix green: 32 passing checks across Python 3.10–3.14 on
  Linux / macOS / Windows


operdeck:
Good call on removing the Java-in-Python stuff for the most part. Also helpful to beef up the AGENTS.md file to help make future code more in line with the rest of pdstools. Nice some of this moved to Ruff instead of hoping for LLM magic.

jmelburg (approved):

PR #914, opened 2026-07-29
fix(report_utils): inline local JS so CDN-mode HTML is a single file

StijnKas:
CDN-mode HTML reports (full_embed=False) were emitted as a zip rather than an HTML file. Quarto leaves its local JavaScript in a companion <basename>_files/ folder, and bundle_quarto_resources wraps the pair into <basename>.zip so they travel together. Health Check therefore returned HealthCheck.zip, and callers expecting an HTML file broke.

#679 already established the fix for the CSS half of this problem; it just stopped short of JavaScript. Extend the same approach:

- _inline_css now rewrites relative url(...) references inside the CSS it inlines into data URIs. Lifting CSS text into a <style> block breaks those paths, which orphaned the bootstrap-icons font.
- _inline_js gives local <script src> tags the same treatment. A literal "</script>" in the source is escaped so it cannot close the tag early. Remote references are deliberately preserved, so Plotly and MathJax keep loading from their CDNs.
- inline_local_assets composes the two, and drop_inlined_resources removes the companion folder once nothing references it. Reports that reference an asset type we do not inline (model reports embed figure-html PNGs) keep their folder, so the zip fallback still applies and bundle_quarto_resources is unchanged.

This matters beyond the extension: full_embed=True is the only other route to a standalone report, and it needs esbuild, which hardened images strip for CVE reasons. Inlining gets there without it.

Verified end to end against cdh_sample: health_check() now returns a single 1.8MB HealthCheck.html with no companion folder, all 23 Plotly figures render, and bootstrap/tippy/anchor/clipboard all initialise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,16 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- CDN-mode HTML reports (``full_embed=False``) are now genuinely
+  self-contained single files. Quarto's local JavaScript assets are inlined
+  alongside the CSS that #679 already handled, and fonts/images referenced
+  from within that CSS are embedded as data URIs. The now-unreferenced
+  ``<basename>_files/`` folder is dropped, so ``bundle_quarto_resources``
+  no longer has a resources folder to zip and the Health Check returns
+  ``HealthCheck.html`` rather than ``HealthCheck.zip``. This gives an
+  esbuild-free path to a standalone report, which matters for hardened
+  images that strip the esbuild binary for CVE reasons. Remote CDN
+  references (Plotly, MathJax) are deliberately left untouched.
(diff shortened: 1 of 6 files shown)


operdeck:
Side effect from a half-baked feature that I did a while back to zip up the resources in CDN mode. That never really worked because CDN wasn't implemented fully, but now that got fixed a few commits ago, the reports started showing as zips. This PR instead puts the resources in the HTML so that can be copied around safely and just by itself. Zipping up could be an option for a later day and still is a fall back in case quarto spits out currently unknown resources.

operdeck (commented):
I do not like post-processing HTML because that is fragile by definition, but we need to for the time being because of the removal of "esbuild" in the DJS job (for (perceived or real) vulnerability issues). To give more confidence I strengthened the test cases around this.

operdeck (approved):

PR #916, opened 2026-07-31
docs: refine AGB explained framing

operdeck:
Few clarifications to align with Shane's deck on AGB


examples/articles/AGBExplained.ipynb
```diff
@@ -579,8 +591,8 @@
     "\n",
     "### Predictors pre-processing\n",
     "\n",
-    "- **Symbolic predictors** use a custom encoding that allows for more granular bins than the classic Naïve Bayes algorithm.\n",
-    "- **Numeric predictors** are pre-processed using **percentile streaming**, which is robust against extreme outliers.\n",
+    "- **Symbolic predictors** are encoded into a bounded bin space. This gives AGB more usable symbolic resolution than the classic Naïve Bayes algorithm while keeping very high-cardinality values from exploding the model; rare values may share bins.\n",
```
EHenryPega, line 594: This is slightly misleading. New distinct values are encoded into a bounded bin space, after which we use hashing to allocate to a pre-existing bin. We do not try to guide rare values into bins with other rare values. Rare values are more likely to share bins in that they are more likely to arrive once all the bins are full. However they have an equal chance of being assigned to all bins.

Suggest something more like: `Symbolic predictors are encoded into a bounded bin space. The first distinct values seen each receive a dedicated bin. Subsequent unseen values are hash-mapped into existing bins, bounding model size regardless of cardinality. This may cause late-arriving — often rare — values to share bins.`

EHenryPega (approved):
Approved with one minor suggestion. Will leave it to your discretion.

examples/articles/AGBExplained.ipynb
operdeck: Thanks, good nuance

jmelburg (approved):

PR #921, opened 2026-08-03
001 add auc confidence intervals

operdeck:
Adding a AUC confidence interval to the ADM NB model reports


StijnKas (commented):
All of the CI calculations are Series based... means it all has to be in memory, can't be lazily evaluated and can't drop to the Polars streaming engine? Can we not rebuild them as expressions?

operdeck:
> All of the CI calculations are Series based... means it all has to be in memory, can't be lazily evaluated and can't drop to the Polars streaming engine? Can we not rebuild them as expressions?

Agent says no: your intuition is correct that the current functions are not fully streamable, but the fundamental reason is the per-model grouped statistical calculation, not just the use of pl.Series.

StijnKas (approved):

PR #934, opened 2026-08-06
Proving Otto Wrong Again

StijnKas:
See https://github.com/pegasystems/pega-datascientist-tools/pull/921#issuecomment-5192341086 / #921 

Replace the grouped Python UDFs used by `active_ranges()` with native Polars
expressions.

The implementation keeps classifier bins in long form and calculates:

- active bin boundaries;
- full-range and active-range AUC;
- grouped DeLong variance;
- confidence interval bounds and availability metadata.

This removes all `map_groups` calls and `python_udf` nodes from the
`active_ranges()` logical plan. It also calculates the confidence interval once
per model rather than invoking the complete calculation separately for each
output field.

## Performance

Median end-to-end `active_ranges().collect()` timings on Polars 1.42.1:

| Models | Original UDF implementation | Native expressions | Speedup |
|---:|---:|---:|---:|
| 4 | 14.10 ms | 6.36 ms | 2.2x |
| 77 (Pega 7 fixture) | 162.67 ms | 10.62 ms | 15.3x |
| 200 | 473.30 ms | 12.71 ms | 37.2x |
| 400 | 956.59 ms | 17.68 ms | 54.1x |

The outputs match the original implementation within `1e-12`.

## Validation

- Added exact-value parity coverage between the Series and expression
  implementations.
- Added a regression check that the logical plan contains no `python_udf`.
- Added default-engine versus streaming-engine result parity coverage.
- Existing active-range edge cases remain covered.

Polars 1.42.1 still executes grouped rank and cumulative-window operations using
in-memory physical-plan nodes. The calculation is therefore expression-only and
compatible with streaming collection, but it is not yet fully out-of-core.

### DIFF
--- a/python/pdstools/adm/ADMDatamart.py
+++ b/python/pdstools/adm/ADMDatamart.py
@@ -18,6 +18,7 @@ from ..pega_io.File import read_dataflow_output, read_ds_export
 from ..utils import cdh_utils
 from ..utils.cdh_utils import _polars_capitalize
 from ..utils.cdh_utils._io import _DATABRICKS_MODEL_SNAPSHOTS_COLUMNS
+from ..utils.cdh_utils._metrics import _auc_ci_from_binned_rows
 from . import Schema
 from .Aggregates import Aggregates
 from .Analysis import Analysis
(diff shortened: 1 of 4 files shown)


operdeck (approved):
Tsss. Hoezo again?

